PR #87, opened 2016-08-02
CMake code to find CGAL and configure TECA for use with it

burlen:
WIP: for use on Grzegorz topological skeleton work


burlen:
this is long dead code, it will not likely be merged, but save the branch because CGAL has some interesting capabilities

PR #92, opened 2016-08-11
WIP -- first pass at jet stream detection algorithm

burlen:
a scratch space for sharing our hacks on Jiabin's algorithm. this should not be merged.


burlen:
Jiabin is no longer with us, this code is pretty basic example of coding up a new algorithm in Python, we have examples like this in the user guide

PR #98, opened 2016-11-28
Programmable reduce

burlen:
make it easier to implement different reductions in out map-reduce framework. users can provide a callback that implements the reduction from 2 datasets to 1.


burlen:
this needs a test and python bindings

burlen:
added Python bindings

PR #135, opened 2017-09-29
Ar hackathon

burlen:
this work is the collaboration of Hari, Travis and my self on AR detectors.

Everything including and below 5fb96e3 should not be merged with this branch. these are all features on other branches that will be merged there.

### DIFF
--- a/alg/teca_ar_rutz.py
+++ b/alg/teca_ar_rutz.py
@@ -2,51 +2,6 @@ import numpy as np
 import teca_py
 import sys
 
-def calculate_pressure_integral(integrand, pressure):
-    """ Calculates the vertical integral (in pressure coordinates) of an array
-    
-        input:
-        ------
-            integrand     : the quantity to integrate.  The vertical dimension is assumed to be the first index.
-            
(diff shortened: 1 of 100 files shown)


alg/teca_ar_rutz.py
```diff
@@ -26,20 +71,56 @@ def update(self):
...
+            md_in[0].append('variables', 'IVT')
+
             return md_in
+        
         return report_callback
 
     def get_request_callback(self):
+        """ Returns a proper TECA request callback function """
         def request_callback(port, md_in, req_in):
-            # pass the request through
```
burlen: We probably will need to allow user to change the name of these variables. Pressure field may be stored as "pressure" in out simulation output and "PSL" in another.

alg/teca_ar_rutz.py
```diff
@@ -26,20 +71,56 @@ def update(self):
...
+            """ Requests the variables needed to find ARs """
+            # add the name of arrays that we need to find ARs
+            req_in['arrays'] = ['QV','u','v','pressure','lat','lon']
```
burlen: We probably will need to allow user to change the name of these variables. Pressure field may be stored as "pressure" in out simulation output and "PSL" in another.

PR #144, opened 2018-11-29
Anaconda compat

burlen:
* thread-safe without HDF5 thread-safe
* switch from Python 2 to Python 3


burlen:
@taobrienlbl this should let us deploy through anaconda (via PyPi). It does move us fully to Python 3 , so Python 2 is not supported after this. I need some test data in netcdf 4 hdf5 format, do you have any?

PR #146, opened 2019-01-10
WIP - work for probabilistic AR detector pipeline

taobrienlbl:
img_0488
Shows shows the pipeline for 1 instance of the probabilistic AR detector running on a single timestep, including which components need which pieces of the request.  Components surrounded in red already exist as TECA algorithms.  The rest need to be implemented.

The naming convention for the AR detector parameters will be: `percentile`, `minimum_area`, and `filter_lat_width`.

Currently, the output of the connected component routine is a 2D field with pixels set to 0 or to the ID of the connected-component to which they belong.  An alternative representation would be as a list of lists: the inner list corresponds to the array indices associated with a given connected-component, and the outer list is a list of these lists of pixels.  See [`RectilinearConnectedComponentSearcher::do_connected_component_search_with_sum_condition()`](https://bitbucket.org/taobrienlbl/bayesian_ar_detector/src/master/floodfillsearch/src/connected_component.cxx) (line 601).


taobrienlbl:
Regarding the suggestion below:

>  An alternative representation would be as a list of lists: the inner list corresponds to the array indices associated with a given connected-component, and the outer list is a list of these lists of pixels.

In the current design, the reduction stage averages the AR identifications over all parameter combinations.  This requires averaging over a bunch of 2D fields that are binary masks: 0=no AR, 1=AR.  If we go with the list-based representation above, then we would need to re-construct these 2D binary mask fields from the list of lists prior to the averaging step; this would be an additional TECA algorithm.  This wouldn't be a huge deal, but it sounds like less work at the moment to stick with the current scheme.

taobrienlbl:
From zoom conversation 3/8/19, TODO:
* Write regression test - 2D array w/ integer values increasing  from upper left corner to lower right corner; calculate which gridcells should be above
* remove segmentation from connected component algorithm (take binary mask as input)?

burlen:
see also #188

burlen:
#188 was intended to replace this mr. #188 was created because github doesn't allow one to change the branch associated with a mr, and I misspelled the branch name here.

PR #166, opened 2019-03-13
Corrected the regression test

elbashandy:
Added a check for hwhm index to match 0.5 in the output array


burlen:
:+1:  looks good

PR #178, opened 2019-03-22
generalize the executive to process datasets not grouped by time step

burlen:
originally the executuve relied on 2 keys

1. number_of_time_steps, to know how many requests should be generated
   to process all the available data

2. time_step, the key that the data source recognizes when the executive
   needs specific subset of the data

There was the assumption that datasets are partitioned by time step.
The executive would make number_of_time_steps requests across all MPI
rank, each request would identify a specific time_step through that key.

Until now when we had datasets that could not be partitioned by time
steps, for example processing tracks, we'd populate the these keys and
ignore the fact they were not really time steps.

this patch generalizes the executive to process datasets partitioned by
other indicies than time steps. In the generalized approach a data
source must name the keys that are used to control request generation.

1. index_initializer_key, a key that names another key which contains
   the number of requests needed to process the available data

2. index_request_key, a key that names the key that will be used by
   the executive when making a specific request.

Data sources such as readers, now must populate these keys. The
executive will use the names it finds to initialize the set of requests
needed to process the available data.


burlen:
i just pushed again because I had not based my branch on the latest master. no new changes

burlen:
rebased to get find python fixes, no new changes

PR #179, opened 2019-03-22
Fix find python

burlen:
this should fix the 21 failing tests on Mac travis, which are caused by cmake finding 2 different versions of python. As part of our transition to python 3, I removed a bunch of the code we had to do that to see if cmake had yet solved its python finding issues. it has not. however, in cmake 3.12 and newer there is a new find python module that claims to fix the issue. unfortunately this version of cmake is not available  on some recent distros, such as ubuntu 18.04 and fedora 28


burlen:
I just pushed again because I had not based off latest master

PR #184, opened 2019-03-26
Polished teca_latitude_damper

elbashandy:
Polished teca_latitude_damper by adding get_output_metadata & it's boost options

### DIFF
--- a/alg/teca_latitude_damper.cxx
+++ b/alg/teca_latitude_damper.cxx
@@ -3,9 +3,15 @@
 #include "teca_variant_array.h"
 #include "teca_metadata.h"
 #include "teca_cartesian_mesh.h"
+#include "teca_metadata_util.h"
 
 #include <iostream>
 #include <set>
+
+#if defined(TECA_HAS_BOOST)
+#include <boost/program_options.hpp>
+#endif
+
 #include <complex.h>
 
 using std::cerr;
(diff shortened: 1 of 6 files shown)


alg/teca_latitude_damper.cxx
```diff
@@ -160,6 +220,36 @@ std::vector<teca_metadata> teca_latitude_damper::get_upstream_request(
 
     arrays.insert(damped_vars.begin(), damped_vars.end());
 
+    // Cleaning off the postfix for arrays passed in the pipeline.
```
burlen: looks good. 

this is a common use case. can you throw the code into a utility function? the question is where does it go, maybe `core/teca_metadata_util.{h,cxx}`?

alg/teca_latitude_damper.cxx
```diff
@@ -152,6 +213,16 @@ std::vector<teca_metadata> teca_latitude_damper::get_upstream_request(
 
     arrays.insert(damped_vars.begin(), damped_vars.end());
 
+    // Cleaning off the postfix for arrays passed in the pipeline. 
+    // For ex a down stream could request "foo_damped" then we'd
+    // need to request "foo". also remove "foo_damped" from the
+    // request.
+    std::string var_post_fix = this->variable_post_fix;
```
burlen: Hi Abdel, I'm seeing that we are making a copy of this string in a couple of the changes. it would be better to take a const reference to it, or just use the member variable directly. doing one of those would avoid copying the string.

Issue #202, opened 2019-04-22
TECA python pipeline hanging w/ MPI and set_enable_mpi(0)


taobrienlbl:
In trying to run a TECA pipeline that has MPI disabled (i.e., `teca_table_reduce.set_enable_mpi(0)` has been used), the pipeline freezes when using `mpirun` for more than one task.

```bash
$ mpirun -n 1 python -u TECAMERRAARCounter.py
start TECAMERRAARCounter.count((0.85, 30.0, 1000000000000.0))
STATUS: [0:139992704821056] [/opt/TECA/core/teca_index_reduce.cxx:111 2.2.1-29-ga1b090a]
STATUS: map index decomposition:
0 : 0 - 23
end TECAMERRAARCounter.count((0.85, 30.0, 1000000000000.0))
```

The above `mpirun -n 1` command runs w/o issue.

```bash
$ mpirun -n 2 python -u TECAMERRAARCounter.py
start TECAMERRAARCounter.count((0.85, 30.0, 1000000000000.0))
```

The above `mpirun -n 2` command stalls when the algorithm's `.update()` function is called.  The stall appears to happen early in TECA's pipeline; if I give an invalid regex, the pipeline immediately fails for the `mpirun -n 1` call, but it still stalls for the `-n 2` call.

The script contents follow.

**TECAMERRAARCounter.py:**
```python
import teca
import numpy as np

class TECAComponentCount:
    
    def __init__(self):
         return
        
    def get_execute_callback(self):
        def execute(port, data_in, req):
            # get the input mesh
            in_mesh = teca.as_teca_cartesian_mesh(data_in[0])
            
            ts = in_mesh.get_time_step()
            tsa = teca.teca_variant_array.New(np.array([ts],'int32'))
            
            md = in_mesh.get_metadata()
            try:
                num_ars = len(md['component_area'])
            except:
                num_ars = 0
            
            ncc = teca.teca_variant_array.New(np.array([num_ars],'int32'))
            
            tab = teca.teca_table.New()
            tab.append_column('step', tsa)
            tab.append_column('count', ncc)
            
            return tab
        
        return execute

class TECAMERRAARCounter:
    """ Counts ARs in MERRA output. This is designed to be run repeatedly after updates to the parameter inputs (i.e., MCMC).
    
        This pipeline is based explicitly on the pipeline defined in TECA/alg/teca_bayesian_ar_detect.cxx at revision 2104b37.
        
[...]

Issue #183, opened 2019-03-26
Runtime setting of the MPI communicator?


taobrienlbl:
TECA currently assumes that the MPI communicator is `MPI_COMM_WORLD` (see below output of `grep -r MPI_COMM_WORLD` in 3c0304c).  It could be useful if this were generalized.  This issue is meant to be a starting point for discussing whether we should do this and, if so, how.

For the the training phase of the Bayesian AR detector, the Markov Chain Monte Carlo 
 (MCMC) algorithm that we use [emcee](http://dfm.io/emcee/current/user/advanced/#using-mpi-to-distribute-the-computations), which also uses MPI for parallelization.  This is the phase that generates the set of AR detector parameters that the AR detector will then apply in parallel to output data (as in #146).  The MCMC algorithm will need to run TECA over a large set of timesteps once for each test parameter.  (Note: for the training phase of the detector, TECA has to be run with only one parameter setting at a time, unlike the run phase, where TECA will need to run on an ensemble of parameters simultaneously; so TECA's parallelism in this phase can only be in the time dimension.)  It would be convenient if both the MCMC algorithm and TECA could both make use of MPI simultaneously.

Pseudocode for this MCMC operation follows:

```python
# create a set of starting parameters (theta = (percentile threshold, filter width, minimum area)
# for each parallel MCMC chain
theta = ... # theta has shape [3, num_mcmc_parallel]

# emcee runs the for-loop below in parallel using MPI
for i in range(num_mcmc_parallel):
     # get the parameters for the current MCMC chain
     current_theta = theta[:, i]
     
     # run TECA for the current theta
     teca_output = run_teca(current_theta) # ideally, this would also run with MPI

     # communicate with other MCMC chains to determine the next version of theta for
     # all chains
     theta = emcee_mpi_get_new_theta(teca_output)
```

Theoretically, I think this should be possible if both `emcee` and TECA each use a separate MPI communication group.  `emcee` allows run-time setting of its MPI communication group, so I think if TECA allows this, it would be possible for both to utilize MPI.  I think that this would work as follows:

[...]

PR #203, opened 2019-04-24
make the MPI communciator setable

burlen:
start to work on #183 #201 #202 

when this is ready for review I will remove the WIP from the pull request name


burlen:
gcc in Ubuntu 14.04 is segfaulting when compiling.
http://cdash.hpcvis.com/viewBuildError.php?buildid=4091

taobrienlbl:
@burlen: [CDASH](http://cdash.hpcvis.com/testDetails.php?test=53421&build=4145) shows a test failure for test_binary_segmentation on Ubuntu_14.04_LTS_user_specified_communicator_941954d: `Attempting to use an MPI routine before initializing MPICH`

burlen:
@taobrienlbl I think this is ready for testing and review, would you please test it and see if it resolves your issue?

core/teca_algorithm.cxx
```diff
@@ -121,14 +127,50 @@ teca_algorithm_internals::teca_algorithm_internals()
     name("teca_algorithm"),
     data_cache_size(1),
     modified(1),
-    exec(teca_algorithm_executive::New())
+    exec(teca_algorithm_executive::New()),
+#if defined(TECA_HAS_MPI)
```
taobrienlbl: @burlen - could you please clarify what this block is doing (perhaps clarifying w/ comments in the code)?

Is this setting a default for the communicator?

burlen:
`test_binary_segmentation` is failing on Travis Ubuntu 14.04,
```
Attempting to use an MPI routine before initializing MPICH
```
http://cdash.hpcvis.com/testDetails.php?test=53421&build=4146

This is a serial test and as far as I can tell makes no MPI calls. I have tested locally with Fedora 28 and both OpenMPI and MPICH and am not able to reproduce. I have also made an effort to look at all the code paths through the test and am not seeing anywhere that would use MPI without first checking to see if it was initialized. I could be wrong but this is looking a lot like a bug in the very likely ancient MPICH deployed on Ubuntu 14.04.

taobrienlbl:
@burlen , in addition to the reference to `MPI_COMM_WORLD` that I just commented on for `core/teca_algorithm.cxx`, there are two other places where `MPI_COMM_WORLD` is hard-coded in core TECA code: in `teca_parallel_id.cxx` and `teca_mpi_manager.cxx`.  

```
./core/teca_parallel_id.cxx:22:        MPI_Comm_rank(MPI_COMM_WORLD, &rank);
./core/teca_mpi_manager.cxx:24:    MPI_Comm_rank(MPI_COMM_WORLD, &m_rank);
./core/teca_mpi_manager.cxx:25:    MPI_Comm_size(MPI_COMM_WORLD, &m_size);
```

This might be intentional and necessary, but if so, it isn't clear from looking at the code why.  If it is intentional/necessary, could you please add comments into the code to clarify?

core/teca_algorithm.cxx
burlen: yes, that is where the default communicator is set. If no one calls `set_communicator` then then `MPI_COMM_WORLD` is returned from `get_communicator`. I'll add a comment.

burlen:
> @burlen , in addition to the reference to `MPI_COMM_WORLD` that I just commented on for `core/teca_algorithm.cxx`, there are two other places where `MPI_COMM_WORLD` is hard-coded in core TECA code: in `teca_parallel_id.cxx` and `teca_mpi_manager.cxx`.
> 
> ```
> ./core/teca_parallel_id.cxx:22:        MPI_Comm_rank(MPI_COMM_WORLD, &rank);
> ./core/teca_mpi_manager.cxx:24:    MPI_Comm_rank(MPI_COMM_WORLD, &m_rank);
> ./core/teca_mpi_manager.cxx:25:    MPI_Comm_size(MPI_COMM_WORLD, &m_size);
> ```
> 
> This might be intentional and necessary, but if so, it isn't clear from looking at the code why. If it is intentional/necessary, could you please add comments into the code to clarify?

`teca_parallel_id` should always use the rank in `MPI_COMM_WORLD`. `teca_mpi_manager` is a helper that ensures MPI finalize is called in tests and applications, when errors occur, and I can't think of a use case it would make sense for it not to use `MPI_COMM_WORLD`.

I can add a comment

burlen:
> @burlen: [CDASH](http://cdash.hpcvis.com/testDetails.php?test=53421&build=4145) shows a test failure for test_binary_segmentation on Ubuntu_14.04_LTS_user_specified_communicator_941954d: `Attempting to use an MPI routine before initializing MPICH`

I added a comment about this above. Long story short, as far as I can tell it is a bug in the version of MPICH deployed with  Ubuntu 14.04. I welcome your fresh eyes on it though. Could you take a look and see if I missing something stupid?

taobrienlbl:
@burlen - thanks for clarifying, and comments will be perfect.

Regarding the build failure, would it make sense to see if the test replicates on cori, since that's one of our current target platforms?  The other thing we could consider is that @elbashandy could create a 14.04 docker image in which we could use debugging tools to try to diagnose where in the code this is being triggered and whether it for sure is an MPICH bug.

If you're confident its a bug w/o doing further testing, I'm OK with proceeding; though I'm leaning toward being cautious, since this could also indicate a subtle bug somewhere in the TECA code that could be hard to track down if it pops up later on in an application.

burlen:
> @burlen - thanks for clarifying, and comments will be perfect.
> 
> Regarding the build failure, would it make sense to see if the test replicates on cori, since that's one of our current target platforms? The other thing we could consider is that @elbashandy could create a 14.04 docker image in which we could use debugging tools to try to diagnose where in the code this is being triggered and whether it for sure is an MPICH bug.
> 
> If you're confident its a bug w/o doing further testing, I'm OK with proceeding; though I'm leaning toward being cautious, since this could also indicate a subtle bug somewhere in the TECA code that could be hard to track down if it pops up later on in an application.

My 2 cents is that it is not worth messing any more with 14.04. Abdel's new 18.04 docker based environment should replace 14.04 asap.

I agree that we should test on Cori prior to the release. perhaps we even set up a nightly dashboard submission from cori, so that we are continuously testing on cori.

core/teca_algorithm.h
```diff
@@ -28,6 +29,17 @@ class teca_algorithm : public std::enable_shared_from_this<teca_algorithm>
...
 
+    // set/get the communicator to use at this stage of the pipeline this has
+    // no influence on other stages. We duplicate the passed in communicator
+    // providing an isolated communication space for subsequent operations. By
+    // default the communicator is initialized to MPI_COMM_WORLD, here it is not
+    // duplicated. Thus to put an algorithm into a unique communication space
+    // one should explicitly set a communicator. When an algorithm should not
+    // use MPI, for instance when it is in a nested pipeline, one may set the
+    // communicator to MPI_COMM_SELF.
+    void set_communicator(MPI_Comm comm);
```
burlen: @taobrienlbl comments added

taobrienlbl:
Attempting to re-open the PR, following comments in #183

PR #204, opened 2019-04-25
Route parameters into metadata for AR detector components

### LINKED ISSUE #198: add # of connected components to metadata in bayesian AR reduce stage

taobrienlbl:
This is a work in progress - there are several remaining items to tackle, including improved comments, generalizing the metadata keys in several of the files, and hardcoding of types in a routine (see comment below).

Following #198, this adds the AR detector input parameters into the AR detector pipeline.  The parameter reduction step combines parameters in the metadata.

@burlen or @elbashandy - in the `property_reduce()` subroutine (starting at line 36 of `alg/teca_bayesian_ar_detect.cxx`), I have the parameter type hardcoded as double, which I know isn't good.  Specifically, there are three instances of `std::vector<double>` in the sub. I could use your help to generalize this.
 
Closes #198

### DIFF
--- a/alg/teca_bayesian_ar_detect.cxx
+++ b/alg/teca_bayesian_ar_detect.cxx
@@ -33,6 +33,52 @@
 #include <mpi.h>
 #endif
 
+/*
+   void property_reduce(std::string property_name,
+                        p_teca_dataset dataset_0,
+                        p_teca_dataset dataset_1,
+                        p_teca_cartesian_mesh mesh_out)
+
+    input:
+    ------
+
+      property_name : (std::string) the key of the metadata property in dataset_* on which to do the reduction
+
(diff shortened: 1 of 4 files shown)


burlen:
this all looks fine to me. It's usually OK to hard code the types of metadata, metadata should be small, and we aren't doing heavy calculations on them.

taobrienlbl:
Well that's good to know then: thanks for checking it out.

I'll do a bit of code clean-up, push the changes, and then request the formal review.

burlen:
btw if you are concerned with performance you can use teca_variant_array's directly, these are passed in/out of metadata zero copy. using std::vector results in making copies to/from the underlying variant arrays.

taobrienlbl:
Thanks @burlen -- I'll keep that in mind if performance looks like it might be limited by that.

PR #209, opened 2019-04-26
Closes #185 Supporting multiple files for CF reader

### LINKED ISSUE #185: ability to input a list of files to the cfreader (and date info)
Following a 3/26/19 zoom call, @burlen  and @taobrienlbl agreed that it could be useful if the CF reader could accept something like a vector of filenames as input.  The fallback (I think) would be the regex functionality that exists now.

This is related to #183, where the files over which TECA will loop are a random list of files.  The order of the filenames is important here, since each file corresponds to a count of ARs within that image (which were hand-counted by members of the CASCADE team).  The TECA pipeline that I'll construct for this will ultimately do a reduction that returns a list of the number of ARs that TECA counts within each file; the order of these counts needs to correspond to the order of the counts in the human-count dataset.

For certain types of files (e.g., the MERRA2 ARTMIP dataset - see #177), the time information is not contained. In this case, it could also be useful if the cfreader also accepts udunit-compatible time input (e.g., a vector of 'days since 1979-01-01 00:00:00' values, along with an input string describing the time units of 'days since...').

elbashandy:
Still under test but wanted the code to be reviewed.

### DIFF
--- a/io/teca_cf_reader.cxx
+++ b/io/teca_cf_reader.cxx
@@ -312,6 +312,8 @@ teca_cf_reader::teca_cf_reader() :
     y_axis_variable("lat"),
     z_axis_variable(""),
     t_axis_variable("time"),
+    t_calendar(""),
+    t_units(""),
     periodic_in_x(0),
     periodic_in_y(0),
     periodic_in_z(0),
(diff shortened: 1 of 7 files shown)


io/teca_cf_reader.h
```diff
@@ -107,12 +120,16 @@ class teca_cf_reader : public teca_algorithm
     virtual void set_modified() override;
 
 private:
-    std::string files_regex;
     std::string file_name;
```
burlen: `file_name` should be replaced by the new vector property `file_names`

io/teca_cf_reader.cxx
```diff
@@ -789,6 +815,25 @@ teca_metadata teca_cf_reader::get_output_metadata(
...
+                        "the number of time values provided")
+                    return teca_metadata();
+                }
+
+                p_teca_variant_array_impl<double> t = teca_variant_array_impl<double>::New(n_t_vals);
+                for (size_t i = 0; i < n_t_vals; ++i)
+                {
+                    t->set(i, this->t_values[i]);
+                    step_count.push_back(1);
+                }
```
burlen: FYI, this loop could be replaced with (something like)
```
p_teca_variant_array_impl<double> t = teca_variant_array_impl<double>::New(this->t_values.data(), n_t_vals);
step_count.resize(n_t_vals, 1);
```
Which is potentially more efficient because we'd let the vector handle the copy and allocation simultaneously.

burlen:
I would like to see a test that depends on the time axis

elbashandy:
@burlen Let me know if it's implemented properly

Issue #165, opened 2019-03-04
use latest linux (ubuntu or fedora) image w/ travis ci


burlen:
1. make use of Abdel's recent work on setting docker images up to configure an image for Travis CI
2. set our travis ci linux build to use it

elbashandy:
Please send me our Travis CI info and access

burlen:
I believe (but could be wrong) that there is no travis ci accounts. it's all via github. and also you don't need to do anything in either github or travis. that part has been set up for quite some time. take a look at the yaml files in the repo.

also see this
https://docs.travis-ci.com/user/docker/

let me know if I'm missing something

burlen:
also setup Python 3 environment, see #205

burlen:
this is getting to be a higher priority, the following looks like a compiler bug in the versions used with 14.04. http://cdash.hpcvis.com/viewBuildError.php?buildid=4009

elbashandy:
Done

PR #211, opened 2019-04-27
Fix negative component areas

### LINKED ISSUE #210: component areas negative when lat coordinate is reversed
When the latitude coordinate is reversed (going from +90 to -90, which is common in reanalyses), the component areas are negative.  The fix appears to be easy: adding `abs()` around the area in the summation.  I'll put in a PR on this momentarily.

@burlen and @elbashandy : I wonder if the test suite should include tests that all algorithms yield good results if latitude coordinates are reversed.

taobrienlbl:
Simple fix - adds `abs()` around area during summation.

Closes #210

### DIFF
--- a/alg/teca_2d_component_area.cxx
+++ b/alg/teca_2d_component_area.cxx
@@ -94,7 +94,7 @@ void component_area(unsigned long nlon, unsigned long nlat,
         unsigned long jj = j*nlon;
         for (unsigned long i = 1; i < nlonm1; ++i)
         {
-            area[labels[jj + i]] += abs(rho_sq_d_theta[i]*d_cos_phi_j);
+            area[labels[jj + i]] += std::abs(rho_sq_d_theta[i]*d_cos_phi_j);
         }
     }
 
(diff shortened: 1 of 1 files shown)


alg/teca_2d_component_area.cxx
```diff
@@ -94,7 +94,7 @@ void component_area(unsigned long nlon, unsigned long nlat,
         unsigned long jj = j*nlon;
         for (unsigned long i = 1; i < nlonm1; ++i)
         {
-            area[labels[jj + i]] += rho_sq_d_theta[i]*d_cos_phi_j;
+            area[labels[jj + i]] += abs(rho_sq_d_theta[i]*d_cos_phi_j);
```
burlen: please use `std::abs`, depending on the header files included there are legacy integer only abs functions from ansi C library.

taobrienlbl:
I just modified to use std::abs

> depending on the header files included there are legacy integer only abs functions from ansi C library.

@burlen and/or @elbashandy - could I please request that our CI tests be augmented to check for build success on platforms with such legacy headers?  It would be best if our automated testing caught such issues.

burlen:
we should be able to improve our test coverage when #165 is complete.

In this case, if you did the headers wrong the build would fail. we are doing the headers correctly here. use of std:: namespace specifier is a one of those best practices that you just do.

Issue #213, opened 2019-04-29
python bindings don't work after pip or setup.py install


taobrienlbl:
When installing TECA via setup.py (either using pip or `python setup.py install`), the python bindings don't appear to be installed properly:

```bash
$ python -c 'import teca; test = teca.teca_cf_reader.New()'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
AttributeError: module 'teca' has no attribute 'teca_cf_reader'
$ python -c 'import teca; help(teca)'
Help on module teca:

NAME
    teca

FILE
    /opt/conda/lib/python3.7/site-packages/teca.py
```

This appears to be because `teca.py` contains
```python
import sys, os
from teca import *
```
The `from teca ...` line is self-referential to `teca.py`, which is likely why it imports nothing(Also, the TECA files were installed directly in the site-packages folder, rather than a teca subfolder, which seems odd.)

In contrast, when TECA is installed directly via CMAKE, the `teca.py` file contains
`from teca_py import *`.

However, if I modify the installed `teca.py` to import from `teca_py` instead, a shared library error results:

```bash
$ python -c 'import teca; test = teca.teca_cf_reader.New()'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/opt/conda/lib/python3.7/site-packages/teca.py", line 2, in <module>
    from teca_py import *
  File "/opt/conda/lib/python3.7/site-packages/teca_py.py", line 33, in <module>
    import _teca_py
ImportError: libteca_alg.so: cannot open shared object file: No such file or directory
```

If I set `LD_LIBRARY_PATH` accordingly (but see note below), I get a new error:
```bash
$ python -c 'import teca; test = teca.teca_cf_reader.New()'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/opt/conda/lib/python3.7/site-packages/teca.py", line 2, in <module>
    from teca_py import *
  File "/opt/conda/lib/python3.7/site-packages/teca_py.py", line 12140, in <module>
    from teca_tc_stats import *
ModuleNotFoundError: No module named 'teca_tc_stats'
```

This import error occurs because the install resulted in `teca_tc_stats.py` being installed in `/.../site-packages/lib`.  If I set `PYTHONPATH` accordingly, the `python -c 'import teca; test = teca.teca_cf_reader.New()'` command then works as expected.

**Thoughts:**

[...]

PR #214, opened 2019-04-30
Fixes for pythonpath and rpath distutils install issues

taobrienlbl:
Modifications to address setup.py install issues described in #213 .

The most recent commit fixes the rpath and pythonpath issues.  But when installing via pip (`pip install TECA/`, where `TECA/` is a local TECA clone), the files are still installed directly in the base directory of site-packages.  However, when running `python setup.py install`, the files are installed in a subdirectory of site-packages (like `teca-2.2.1-py3.6-linux-x86_64.egg`).  This should be addressed before this PR is merged.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -37,6 +37,9 @@ class CMakeBuild(build_ext):
 
         # this is where setuptools will look for our files to install
         extdir = os.path.abspath(os.path.dirname(self.get_ext_fullpath(ext.name)))
+        # put verything in a directory below this for pip, if we don't do so
+        # pip puts it all right in site-packages making a big fat mess
+        extdir = os.path.join(extdir, 'teca')
 
         # set some flags for the cmake command line configuring this build
         # specifically we need to put the build where setuptools can find it
@@ -83,14 +86,13 @@ class CMakeBuild(build_ext):
             sys.stderr.write('\n')
             raise
 
-        # generate on the fly the top level python module that can be imported from the
-        # setuptools install
-        f = open(os.path.join(extdir, 'teca.py'), 'w')
+        # because we are putting everything in a subdirectory (see comment above re: pip)
+        # write the init file that actually does the loading sets up the paths correctly
+        f = open(os.path.join(extdir, '__init__.py'), 'w')
         f.write('import sys, os\n')
-        f.write('sys.path.append(os.path.join(os.path.dirname(os.path.abspath(__file__)),"lib"))\n')
-        f.write('from teca_py import *\n')
+        f.write('sys.path.append(os.path.dirname(os.path.abspath(__file__)))\n')
+        f.write('from teca.teca_py import *\n')
         f.close()
-
 
 with open("README.md", "r") as fh:
     long_description = fh.read()
--- a/setup.py
+++ b/setup.py
@@ -87,7 +87,8 @@ class CMakeBuild(build_ext):
         # setuptools install
         f = open(os.path.join(extdir, 'teca.py'), 'w')
         f.write('import sys, os\n')
-        f.write('from teca import *\n')
+        f.write('sys.path.append(os.path.join(os.path.dirname(os.path.abspath(__file__)),"lib"))\n')
+        f.write('from teca_py import *\n')
         f.close()
 
 
--- a/alg/CMakeLists.txt
+++ b/alg/CMakeLists.txt
@@ -46,7 +46,7 @@ set(teca_alg_cxx_srcs
     teca_dataset_diff.cxx
     )
 
-teca_py_install(lib
+teca_py_install(${LIB_PREFIX}
     teca_tc_stats.py
     teca_tc_wind_radii_stats.py
     teca_tc_trajectory_scalars.py
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -19,8 +19,14 @@ if (TECA_HAS_PYTHON)
 
     PYTHON_ADD_MODULE(_teca_py ${CMAKE_CURRENT_BINARY_DIR}/teca_py.cxx)
 
-    # make the rpath relative so that moving the so file doesn't break the link
-    SET_TARGET_PROPERTIES(_teca_py PROPERTIES INSTALL_RPATH "\$ORIGIN/./")
+    # include the directory containing _teca_py in the rpath
+    # this enables one not to have to set (DY)LD_LIBRARY_PATH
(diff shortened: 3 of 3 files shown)


setup.py
```diff
@@ -87,7 +87,8 @@ def build_extension(self, ext):
         # setuptools install
         f = open(os.path.join(extdir, 'teca.py'), 'w')
         f.write('import sys, os\n')
-        f.write('from teca import *\n')
+        f.write('sys.path.append(os.path.join(os.path.dirname(os.path.abspath(__file__)),"lib"))\n')
+        f.write('from teca_py import *\n')
```
burlen: are you trying to run out of the build itself? I think that's what __file__ here will point to. I suspect that this would break the install?

burlen:
this one needs thorough testing. I'll try to run it through it's paces asap.

burlen:
I've only tested use with pip on PyPi repo. All local testing has bee n done with
```
python setup.py install
```
How common is running pip on a source tree?

setup.py
taobrienlbl: No, since `__file__` is within quotes here, it written to `teca.py` (and not resolved during the setup.py process, which I think is what you are implying), this will refer to the file path of `teca.py` once imported: e.g., as in `import teca; print(teca.__file__)`.  So it will refer to the installed path.  This is intended to add the `lib` subdirectory of the TECA install to the path, which is needed since `teca_py.py` imports things like `teca_tc_stats.py`, which exist in lib but aren't part of the path.

taobrienlbl:
pip installing from a source tree isn't super common, but it is used to install a development version of the package while still maintaining the ability to uninstall with pip.  The other option is directly from git: `pip install git+https://github.com/LBL-EESA/TECA.git`.

AFAIK, both the source and git options mimic what installing from PyPi would do: just without the need to upload to PyPi.

setup.py
burlen: OK, got it. I hadn't taken account of the quotes.  I vaguely recall that the solution as implemented was taken directly from the Python packaging documentation. I'll try to find it. Also, I'm very curious as to why all this works for me and not you.

burlen:
> AFAIK, both the source and git options mimic what installing from PyPi would do: just without the need to upload to PyPi.

hmm, you may be right but I do wonder about that, because there is a special process to prepare the upload to PyPi.
```
python3 setup.py sdist
python3 -m twine upload --repository-url https://test.pypi.org/legacy/ dist/*
```

setup.py
taobrienlbl: Perhaps we should have a working (virtual w/ screenshare?) session in the next week or so.

taobrienlbl: Btw when you say 'this works', are you referring to the set of issues described in #213?

burlen: Yup, I was thinking the same thing. A face to face would be the best way to proceed.

`python setup.py install` run from a clone has been working for me on Fedora 28, Ubuntu 18.04, and Mac OS latest. I had been using virtual env's to keep the installs out of the system places.

I believe `$ORIGIN` is a Linux specific trick, and while Mac OS has it's own mechanisms, it would be best to not go into platform specific tricks if we can avoid them.

I would like to, understand what is different about your workflow/environment and mine, and test this on Fedora, Ubuntu and Mac, and in PyPi before merging it.

https://gitlab.kitware.com/cmake/community/wikis/doc/cmake/RPATH-handling
http://man7.org/linux/man-pages/man8/ld.so.8.html
https://packaging.python.org/guides/packaging-namespace-packages/

burlen: I see the issue now. What I had initially done never worked. my testing was not thorough enough, I had been only testing `from teca import *`, which does succeed but nothing is imported. At one point I had a solution similar to yours manipulating paths in the generated file, without using `RPATH`, but I backed it out after being able to import without it.

burlen:
@taobrienlbl I tested your change on Ubuntu 18.04, via `python3 setup.py install`, and it solved the issues there.  Also, It turns out that the pure python algorithms were not being installed to consistently. I committed a fix for that so that we don't need to add `lib` to the python path. I backed that part of your change out, and tested again on Ubuntu 18.04.

TODO:
1. Test the `RPATH` manipulation, Superbuild, and Mac OS. Mac OS may need a different solution. See https://gitlab.kitware.com/cmake/community/wikis/doc/cmake/RPATH-handling
2. consider using `dlopen` from our Python bootstrap code. Would this be any better than manipulation of `RPATH`?
3. With `pip install teca` our libraries are installed in `site-packages` . With `python setup.py install` they end up in `site-packages/teca-2.2.1-py3.6-linux-x86_64.egg`. The latter is probably where they should be? With `pip` the following code is yielding different path `extdir = os.path.abspath(os.path.dirname(self.get_ext_fullpath(ext.name)))` we'll need to find something that works for both `pip install` and `python setup.py`

setup.py
taobrienlbl: okay, glad to confirm that we're now on the same page wrt the issue.

taobrienlbl:
@burlen:

Agreed on the RPATH manipulation; it almost certainly needs a different approach for mac.

I like the idea of attempting to resolve the shared library from within python itself, since my gut feeling is that the RPATH solution I introduced might be a fragile solution.

And agree on your point 3, though I haven't come up with any ideas on that yet.

burlen:
@taobrienlbl  Today I made some tweaks for Mac OS and to make sure when we use pip things don't land directly in `site-packages` making a mess. 
Could you test these changes?

I think we've solved all the issues identified. I'm inclined to merge this, cut a bug fix release with just this change, and push the new release onto PyPi. What do you think?

fyi, watch out because I rebased and thus you'll need to do something like
```
git fetch
git checkout pip_install_fix
git reset --hard origin/pip_install_fix
```  
which would blow away any local changes you had. sorry about that.

taobrienlbl:
@burlen - nice!  I'm testing now, but I'm sure you fixed it.  Assuming it pip-installs w/o issue, I'll go ahead and approve.  Yes, definitely I think this warrants a bug fix release and push to pypi.

taobrienlbl:
Yes, installs via pip perfectly!  And d'oh--this is my PR--I can't approve: you'll need to.  Have a good weekend!

burlen:
OK, will approve it. You deserve all the credit on this one, nice work.

burlen:
@taobrienlbl I merged this because I wanted to get it in before the MPI communicator merge

PR #222, opened 2019-05-02
Building TECA on Ubuntu 18:04 on TravisCI through Docker

elbashandy:


test/CMakeLists.txt
```diff
@@ -148,15 +148,15 @@ teca_add_test(test_descriptive_statistics_threads
 
 teca_add_test(test_descriptive_statistics_mpi
     LIBS teca_core teca_data teca_io teca_alg ${teca_test_link}
-    COMMAND ${MPIEXEC} -n ${TECA_TEST_CORES} test_descriptive_statistics
+    COMMAND ${MPIEXEC} -n ${TECA_TEST_CORES} --oversubscribe test_descriptive_statistics
```
burlen: I think this is an open-mpi specific option, and it may not work to have it hard coded. But before going there, a question.

how many cores are there on these systems? Assuming its not a 1 core system, we might want avoid over subscription. 

You can control the number of cores the tests run over by adding `-DTECA_TEST_CORES=N` where N > 1 to the cmake command line. In the current setup we had been using `TECA_TEST_CORES=2`, this is set in `test/travis_ci/ctest+linux.cmake`

.travis.yml
```diff
@@ -2,22 +2,24 @@
...
-  - gcc
 
-install: ./test/travis_ci/install_$TRAVIS_OS_NAME.sh
-
-env:
-  - BUILD_TYPE=Debug
-#  - BUILD_TYPE=Release
-
-script: ./test/travis_ci/ctest_$TRAVIS_OS_NAME.sh
+script: ctest -I -V
```
burlen: I think we'll need to pass the cmake configs to ctest. These have ctest options that get the branch name into the web UI, among other features. I think you'll need to modify some parts of those scripts, checkout `test/travis_ci/ctest_linux.cmake` and `test/travis_ci/ctest_osx.cmake`

.travis.yml
```diff
@@ -2,22 +2,24 @@
 language: c++
 sudo: required
 
-os:
-  - linux
-  - osx
```
burlen: is there a way to keep osx running as it has been and have the new docker based way for linux?

elbashandy:
@burlen This is ready now, the only thing that's missing is the configure info on cdash. Not sure why this is happening despite the fact that Configure.xml is being uploaded normally.

burlen:
@elbashandy  If you feel like it, take a break from this one. When I have a minute I'll try to see if I can spot the issue

test/travis_ci/docker/ctest_linux.cmake
```diff
@@ -0,0 +1,33 @@
...
+set(CTEST_NOTES_FILES)
+set(ALWAYS_UPDATE ON)
+set(INCREMENTAL_BUILD OFF)
+set(DASHBOARD_TRACK "Continuous")
+set(DATA_ROOT "${CTEST_DASHBOARD_ROOT}/TECA_data")
+set(CTEST_SOURCE_DIRECTORY "$ENV{TECA_DIR}")
+set(CTEST_BINARY_DIRECTORY "${CTEST_DASHBOARD_ROOT}/build")
+set(CTEST_TEST_TIMEOUT 600)
+set(INITIAL_CACHE
+"CMAKE_C_COMPILER=$ENV{CC}
```
burlen: get rid of these,  they were added to work around ubuntu 14.04 old compiler

test/travis_ci/docker/ctest_linux.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+set -v
+export DASHROOT=`pwd`
+export CC=`which gcc`
```
burlen: get rid of these

PR #229, opened 2019-05-07
thread_pool don't use MPI before initialization

burlen:


burlen:
@taobrienlbl Hi Travis, could you give this a review? I think it solves a couple of the issues you reported

Issue #230, opened 2019-05-07
Add test for setting MPI communicator from python


taobrienlbl:
Could be based off the test code in #203 and #202.

PR #231, opened 2019-05-07
more executive generalization

burlen:
* writers use index in file names
* data sources pass index in dataset metadata
* reductions isolate executive controls from pipeline below/above the reduction
* dataset diff mediates between pipeline below and the 2 pipelines above
* a handful of bug fixes and small features

### DIFF
--- a/alg/teca_bayesian_ar_detect.cxx
+++ b/alg/teca_bayesian_ar_detect.cxx
@@ -16,6 +16,7 @@
 #include "teca_programmable_algorithm.h"
 #include "teca_programmable_reduce.h"
 #include "teca_dataset_capture.h"
+#include "teca_index_executive.h"
 #include "teca_mpi.h"
 
 #include <algorithm>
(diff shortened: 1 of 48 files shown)


burlen:
closes #236,#230

burlen:
@taobrienlbl could you help me with a review? this started as a few small features but evolved into a big change to update writers to use the new more general map index keys instead of hard coding for the  time_step key.

taobrienlbl:
@burlen - I've reviewed the best I can, and it looks good over all.  That said, there are a lot of code modifications to parts of TECA where I'm not yet familiar with the purpose of the code.  So I'm not sure how effective my review is.

Also, because there are so many different modifications, I've had a hard time wrapping my head around the implications of all of these changes.

Perhaps, given the scope of this PR, it would make sense to get additional eyes on this; that could help with covering the parts that I'm not able to adequately review.

burlen:
yes, this is a big set of change, and complex. I'm afraid beside myself you know teca internals better than anyone. I understand that this one may not be realistically review-able. I tested on your script and it still worked.

PR #238, opened 2019-05-21
Coordinate system normalization

burlen:
* adds machinery to deal with incompatible coordinate systems such as inverted coordinate axes.
* adds support for linear interpolation in the re-meshing operation
* adds reader/writer for TECA's internal mesh format, this enables mesh based diff in regression tests
* adds a mesh source to simplify testing

### DIFF
--- a/alg/teca_normalize_coordinates.cxx
+++ b/alg/teca_normalize_coordinates.cxx
@@ -1,0 +1,451 @@
+#include "teca_normalize_coordinates.h"
+
+#include "teca_cartesian_mesh.h"
+#include "teca_array_collection.h"
+#include "teca_variant_array.h"
+#include "teca_metadata.h"
+#include "teca_coordinate_util.h"
+
+#include <algorithm>
+#include <iostream>
+#include <string>
+#include <set>
+#include <cmath>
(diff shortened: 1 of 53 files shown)


alg/teca_normalize_coordinates.cxx
```diff
@@ -0,0 +1,440 @@
...
+p_teca_variant_array teca_normalize_coordinates::internals_t::normalize_axis(
+    const const_p_teca_variant_array &x)
+{
+    unsigned long nx = x->size();
+    unsigned long x1 = nx - 1;
+
+    NESTED_TEMPLATE_DISPATCH(const teca_variant_array_impl,
+        x.get(), _C,
+
+        const NT_C *px = dynamic_cast<TT_C*>(x.get())->get();
```
taobrienlbl: Lines 55-66 are a sample block of code where it would be useful to add a bit of commentary indicating the function of the code block.  On reading through the code, I think that you are checking whether the coordinate needs to be reversed to be ordered negative->positive, but I'm not sure.

burlen:
@taobrienlbl added comments

PR #245, opened 2019-05-22
Pdi

burlen:
*  generate Python exceptions from C++
* pass numpy scalars through the wrapped api
* add PDI calcs & plots, added to app, and tests
* bug fix for serial table reader/ writer


burlen:
@taobrienlbl would you check this one out when you get a chance?

burlen:
@taobrienlbl thank you for looking at it! I rebased and fixed conflicts, added some documentation. done.

PR #250, opened 2019-05-31
Testing TECA on Ubuntu 18:04 & Fedora 28 using Docker on Travis-CI

elbashandy:


test/travis_ci/docker/ubuntu_18_04/install_ubuntu.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+set -v
+
+# setup repo with recent gcc versions
+add-apt-repository -y ppa:ubuntu-toolchain-r/test
```
burlen: this is cruft from 14.04. there should be no need to use ppa's with 18.04 as the default package versions should be recent enough. can you please use apt-get install for these packages now?

.travis.yml
```diff
@@ -2,22 +2,76 @@
...
+    export IMAGE_VERSION=$IMAGE_VERSION &&
+    $TECA_DIR/test/travis_ci/docker/ctest_linux.sh";
+    fi
 
 after_failure:
-  - cat build/Testing/Temporary/LastTest.log
+  - >
+    if [[ "$TRAVIS_OS_NAME" != "linux" ]];
+    then
+    cat build/Testing/Temporary/LastTest.log;
```
burlen: fix indentation, here and elsewhere shell script should be indented with 4 spaces

test/travis_ci/ctest_linux.cmake
```diff
@@ -30,4 +30,4 @@ ctest_read_custom_files("${CTEST_BINARY_DIRECTORY}")
 ctest_build()
 ctest_submit(PARTS Update Configure Build Notes RETRY_DELAY 15 RETRY_COUNT 10)
 ctest_test()
-ctest_submit(PARTS Test RETRY_DELAY 15 RETRY_COUNT 10)
+ctest_submit(PARTS Test RETRY_DELAY 15 RETRY_COUNT 10)
```
burlen: please, leave this change out. it doesn't add anything.

test/travis_ci/docker/ubuntu_18_04/install_ubuntu.sh
```diff
@@ -0,0 +1,32 @@
...
+add-apt-repository -y ppa:teward/swig3.0
+#add-apt-repository -y ppa:george-edison55/cmake-3.x
+
+# suck in package lists
+apt-get update -qq
+
+# install deps
+# use PIP for Python packages
+apt-get install -qq -y git-core gcc-5 g++-5 gfortran-5 cmake swig3.0 \
```
burlen: more cruft from 14.04. please get rid of version numbers  on packages. should not be necessary as 18.04 has new enough versions, where 14.04 did not (eg. gcc 4 didn't have full c++11 implementation hence, we needed 5)

test/travis_ci/docker/ubuntu_18_04/install_ubuntu.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+set -v
+
+# setup repo with recent gcc versions
+add-apt-repository -y ppa:ubuntu-toolchain-r/test
+add-apt-repository -y ppa:teward/swig3.0
+#add-apt-repository -y ppa:george-edison55/cmake-3.x
+
+# suck in package lists
+apt-get update -qq
```
burlen: is this needed once we get rid of ppa's? if not then get rid of this line too

test/travis_ci/docker/ctest_linux.cmake
```diff
@@ -0,0 +1,30 @@
...
+set(INCREMENTAL_BUILD OFF)
+set(DASHBOARD_TRACK "Continuous")
+set(DATA_ROOT "${CTEST_DASHBOARD_ROOT}/TECA_data")
+set(CTEST_SOURCE_DIRECTORY "${CTEST_DASHBOARD_ROOT}")
+set(CTEST_BINARY_DIRECTORY "${CTEST_DASHBOARD_ROOT}/build")
+set(CTEST_TEST_TIMEOUT 600)
+set(INITIAL_CACHE
+"CMAKE_BUILD_TYPE=$ENV{BUILD_TYPE}
+BUILD_TESTING=ON
+TECA_PYTHON_VERSION=2
```
burlen: could we read the python version from an environment variable?

.travis.yml
```diff
@@ -2,22 +2,76 @@
...
 
-dist: trusty # Ubuntu 14.04.
+env:
+  global:
+    - BUILD_TYPE=Debug
+    - TECA_DIR=/travis_teca_dir
+    - TECA_DATA_REVISION=39
+  matrix:
+    - DOCKER_IMAGE=ubuntu IMAGE_VERSION=18.04 SCRIPT_DIR=ubuntu_18_04
+    - DOCKER_IMAGE=fedora IMAGE_VERSION=28 SCRIPT_DIR=fedora_28
```
burlen: could we add a line(s) for Python 2 & 3 runs? see comment below about using environment var for the Python version. I know we said we'd do this in another pr, but it seems almost trivial addition...

.travis.yml
```diff
@@ -2,22 +2,76 @@
...
 os:
   - linux
   - osx
 
-dist: trusty # Ubuntu 14.04.
+env:
+  global:
+    - BUILD_TYPE=Debug
+    - TECA_DIR=/travis_teca_dir
+    - TECA_DATA_REVISION=39
```
burlen: excellent!

test/travis_ci/docker/ubuntu_18_04/install_ubuntu.sh
```diff
@@ -0,0 +1,32 @@
...
+cd libxlsxwriter
+make
+make install
+cd ..
+
+pip install --user numpy mpi4py matplotlib
+
+# install cmake manually because repo/ppa versions are too old
+#wget https://cmake.org/files/v3.5/cmake-3.5.2-Linux-x86_64.tar.gz
+#tar -C /usr -x -z -f cmake-3.5.2-Linux-x86_64.tar.gz --strip-components=1
```
burlen: 14.04 cruft just delete these lines

test/travis_ci/docker/ubuntu_18_04/install_ubuntu.sh
```diff
@@ -0,0 +1,32 @@
...
+
+# install data files.
+svn co svn://missmarple.lbl.gov/work3/teca/TECA_data@${TECA_DATA_REVISION} TECA_data
```
burlen: nice!

test/travis_ci/docker/fedora_28/install_fedora.sh
```diff
@@ -0,0 +1,27 @@
...
+cd libxlsxwriter
+make
+make install
+cd ..
+
+source /usr/share/Modules/init/bash
+
+module load mpi
+
+pip install --user numpy mpi4py matplotlib
```
burlen: question:  why --user? also that might need to be pip3 if we are using Python 3?

.travis.yml
```diff
@@ -2,22 +2,76 @@
 language: c++
 sudo: required
 
+services:
+  - docker
+
 os:
   - linux
```
burlen: question: does this line run 14.04? if so do we still want to be running 14.04? I'm thinking it's more trouble than its worth. If we are getting rid of it we could remove any scripts used only by 14.04.

.travis.yml
elbashandy: Yeah I tried this but that format doesn't work with yaml.  I tried even their [indentation feature](https://yaml.org/spec/1.2/spec.html#id2793979) and it didn't work because it requires all the lines to be equally indented.

test/travis_ci/docker/fedora_28/install_fedora.sh
elbashandy: good point. Are we using matpotlib?

.travis.yml
elbashandy: Docker service is only available on linux (which is 14.04). I can remove the 14:04 scripts responsible for building TECA

burlen: hmm. Sorry but I am confused about your reply. We are talking about the os section are we not?

test/travis_ci/docker/ctest_linux.sh
```diff
@@ -0,0 +1,27 @@
...
+export PYTHONPATH=${DASHROOT}/build/lib
+export LD_LIBRARY_PATH=${DASHROOT}/build/lib
+export MPLBACKEND=Agg
+mkdir build
+cmake --version
+#export TRAVIS_BRANCH=$TRAVIS_BRANCH &&
+#export BUILD_TYPE=$BUILD_TYPE &&
+#export DOCKER_IMAGE=$DOCKER_IMAGE &&
+#export IMAGE_VERSION=$IMAGE_VERSION &&
+#export TECA_PYTHON_VERSION=${TECA_PYTHON_VERSION} &&
```
burlen: could you please clean out the stuff that isn't needed too?

.travis.yml
burlen: OK

burlen: OK. I guess we have to live with it, or move shell code into a script. If you want to go for it! If not, this is fine.

test/travis_ci/docker/fedora_28/install_fedora.sh
burlen: yes

.travis.yml
burlen: huh, I'm confused. I meant to inquire about the os section, you seemed to reply about the services section...

Issue #215, opened 2019-04-30
potential bayesian_ar property reduce crash


burlen:
the new property reduce function needs to handle all the input cases. The cases are both datasets are valid, the left dataset is valid, the right dataset is valid, neither dataset is valid. the current implementation only handles the first case and would crash on the rest.

burlen:
the reduce is currently written so that it is only called on pairs of datasets. Both inputs are always valid.

However this creates potential problem/inconsistency when there is only 1 dataset. This dataset will not be processed by the reduction. the problem comes when the reduction generates something new, removes things, or otherwise transforms the data. it would mean that you get different result with a single dataset. 

admittedly this is a corner case, but I would like to change the implementation to always call the reduce so that this type of filtering can always safely be applied.

burlen:
This can be reproduced by making the parameter table have an odd number of rows. `--parameter_table::number_of_rows 1` note that 1 and >1 are two different cases.

burlen:
@taobrienlbl in reference to your email

burlen:
I believe this is closely related to #305. If that is the case close them both at the same time.

burlen:
tested today with 1 parameter table row verifies that this is no longer and issue.

PR #258, opened 2019-06-17
WIP - bayesian AR detector app

taobrienlbl:
@burlen - I could use your help with this.  I think I've nearly got a working draft of a bayesian AR detector app, but I'm missing something with the cfwriter.  Right now, the cfwriter gets the metadata from the cfreader.  I've got a custom `teca_metata` object constructed for the cf_writer, but I can't seem to figure out how to override the metadata coming from the cf_reader.  Maybe we could have a working session for this?  There are a few other things I could use your input on.


apps/teca_bayesian_ar_detector_parameters.h
```diff
@@ -0,0 +1,6 @@
+using parameter_t = double;
```
taobrienlbl: This file needs to be rewritten to contain actual parameters from the MCMC estimation - this two-parameter header is just for testing.

apps/teca_bayesian_ar_detect.cpp
```diff
@@ -0,0 +1,335 @@
...
+    teca_metadata ar_probability_metadata;
+    
+    // set metadata for the 'ar_probability' variable
+    ar_probability_metadata.set("long_name", std::string("posterior AR flag"));
+    ar_probability_metadata.set("units", std::string("probability"));
+    // load the metadata into the master array
+    cf_metadata.set("ar_probability", ar_probability_metadata);
+    
+    // override the incoming metadata for the executive
+    //cf_writer->get_metadata() = &cf_metadata;
```
taobrienlbl: @burlen - this is where I'm stumped on dealing with the metadata.

apps/teca_bayesian_ar_detect.cpp
```diff
@@ -0,0 +1,335 @@
...
+
+    p_teca_normalize_coordinates sim_coords = teca_normalize_coordinates::New();
+    sim_coords->set_input_connection(cf_reader->get_output_port());
+    
+    /* Load detector parameters into a data table */
+    p_teca_variant_array_impl<parameter_t> min_water_vapor = teca_variant_array_impl<parameter_t>::New(num_ar_detector_parameters);
+    p_teca_variant_array_impl<parameter_t> filter_lat_width = teca_variant_array_impl<parameter_t>::New(num_ar_detector_parameters);
+    p_teca_variant_array_impl<parameter_t> min_area_kmsq = teca_variant_array_impl<parameter_t>::New(num_ar_detector_parameters);
+    
+    for (unsigned long n = 0; n < num_ar_detector_parameters; n++){
```
taobrienlbl: There must be a cleaner way of doing this - @burlen ?

taobrienlbl:
@burlen  - the fixes you pushed in  b1066d0 and a34f938 did the trick; the test case I'm running worked.  Thanks!!

burlen:
@taobrienlbl I made the reorganization we talked about. Now I see why we need to profile! Very slow on large table.

burlen:
@taobrienlbl  I added  `--parameter_table::number_of_rows` command line option to help aid in profiling. The default of -1  serves up the full table, but now I wonder if a smaller size might produce a result that is "good enough" ? & is serving the full table the best choice of a default?

burlen:
wrt to the new parameter table rows control see comments in #215

burlen:
I profiled the code today using callgrind with the above compiler options, it is showing that 70% of the time is spent in `teca_binary_segmentation`. This makes sense, as we are segmenting based on a percentile. Computing the percentile is based on a partial sort which is an expensive operation.

teca_profile

taobrienlbl:
Thanks for reviewing, rebasing, etc. Burlen!  This looks great.  I'll merge now.

PR #260, opened 2019-06-18
teca_bayesian_ar_detect re-use thread pool

burlen:
make the thread pool directly set-able in teca_threaded_algorithm.
create and cache a thread pool for use during teca_bayesian_ar_detect,
this enables the threads to be bound correctly to cores across MPI
ranks, reduces overhead from repeated contruction when parallelizing
over control parameters.


burlen:
@taobrienlbl I think this is OK to go. I tested it with your app and it didn't cause any trouble.

PR #269, opened 2019-07-10
added ability for cfwriter to include date in output filenames

taobrienlbl:
See #270

### DIFF
--- a/data/teca_coordinate_util.cxx
+++ b/data/teca_coordinate_util.cxx
@@ -7,6 +7,9 @@
 
 #include <string>
 #include <cstdio>
+#include <ctime>
+#include <sstream>
+#include <iomanip>
 
 namespace teca_coordinate_util
 {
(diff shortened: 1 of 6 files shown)


taobrienlbl:
@burlen - assuming the TravisCI tests pass, this is ready for you to review.  One question: I added a new utility function `time_step_to_string()` to `data/teca_coordinate_util.cxx`.  Do you think that's the most logical place for that function?

data/teca_coordinate_util.h
```diff
@@ -212,6 +212,12 @@ int index_of(const const_p_teca_cartesian_mesh &mesh, T x, T y, T z,
 int time_step_of(p_teca_double_array time, bool lower,
     const std::string &calendar, const std::string &units,
     const std::string &date, unsigned long &step);
+    
+// given a time value (val), associated time units (units), and calendar (calendar), return a human-readable
+// rendering of the date (date) in a strftime-format (format).  return 0 if successful.
+int time_step_to_string(double val, 
+                        const std::string &calendar,  const std::string &units, const std::string &format,
```
burlen: indent & wrap

data/teca_coordinate_util.cxx
```diff
@@ -59,6 +62,45 @@ int time_step_of(p_teca_double_array time, bool lower,
...
+        return -1;
+    }
+    
+    // fix some of the timedata info to conform to what strftime expects
+    timedata.tm_year -= 1900; // make the year relative to 1900
+    timedata.tm_mon -= 1; // make the months start at 0 instead of 1
+    timedata.tm_sec = (int) std::round(seconds);
+    
+    // use std::put_time to convert the time data to a string
+    std::ostringstream os;
```
burlen: use strftime, given that we don't need the ostream (as discussed I'll make that change before we merge)

burlen:
@taobrienlbl Hi, I cleaned up as we discussed. I noticed that we had Z and not %Z, and when I made that change I found that file names had garbage. Tracked that down to some uninitialized fields of `struct tm`. Lastly I changed the default from `%F-%H%Z` to `%F-%R-%Z`. Is that OK? If you prefer `%H` to `%R` I can change it back, just thought `%R` will be safer in case someone has higher resolution data?

taobrienlbl:
Thanks for doing this.  The use of Z was intentional: it wasn’t meant to have a % in front.  This is to indicate that the ending number is a GMT-0 hour (Z as in Zulu, which is shorthand for GMT-0), which is the prevailing standard for climate output, so it is a reasonable default.  

I’d also like to suggest we keep %H as the default and allow the user to use more complicated time strings if desired.  It will be rare to find a dataset with subhourly resolution, so this default will keep filenames a bit more concise.

taobrienlbl:
Btw, I’ll make that change directly in a few, once back at my desk.

burlen:
Got it! Sorry for messing it up ;-)

taobrienlbl:
no worries, and thanks!  I just squashed the commits, and I'll merge once I the TravisCI tests complete successfully.

burlen:
ok, all done, add my last change and the co-author note. once the tests finish one of us could merge it

PR #273, opened 2019-07-12
WIP - Modifications to allow build w/ intel compilers

taobrienlbl:
Change the way to query if compiler flags have been set or not -- needed for non-gnu build.

Intel compiler complained about the integers having a type already set.
Move declaration up.


burlen:
@ndkeen I did that for you, will merge now

PR #274, opened 2019-08-05
Implemented teca_timer profiler

elbashandy:
@burlen @taobrienlbl this is done. Didn't put the profiler in the right directory. Where do you guys think the profiler should be placed? maybe core/profiler/?


core/teca_algorithm_fwd.h
```diff
@@ -29,6 +29,11 @@ std::shared_ptr<T const> shared_from_this() const   \
 {                                                   \
     return std::static_pointer_cast<T const>(       \
         teca_algorithm::shared_from_this());        \
+}                                                   \
+                                                    \
+std::string get_class_name()                              \
```
burlen: TODO -- return const char * and declare this `const expr`

PR #276, opened 2019-08-12
work around for slow svn on mac os

burlen:
see #275


burlen:
@taobrienlbl I think this is good to go...

PR #279, opened 2019-09-03
make travis build fail if ctest has failing tests

### LINKED ISSUE #280: travis ci builds don't fail when tests fail
I started to poke at this in  #279

1. cmake returns non-zero if any tests fail.
2. what does docker exec return? documentation suggests it should return cmake exit code.

burlen:
* make sure that travis ci reports build as failed if any ctests fail
* update travis ci to use Python 3

resolves #280, #290


burlen:
@elbashandy See anything that could be better? Have any objections?

PR #281, opened 2019-09-04
fix pipeline keys in Bayesian AR algorithm

burlen:


burlen:
@taobrienlbl Hi Travis ,

can you give this a review? It fixes some issues I found with keys in Bayesian AR alg, changes the cf writer default to 1 step per file,  fixes time step subsetting in the cf writer (reported by Noel #283), adds memory use plots to the profiler explorer app, and some python 3 cleanup.

These changes should make things easier for Noel, and the keys fix is something he hasn't encountered yet, but it would be good to get it in before he does.

Burlen

PR #285, opened 2019-09-22
Non spatial io

burlen:
* adds support for writing information arrays(those have no geometric/spatial association)
* uses this in the Bayesian AR detector to capture and output parameter table row, and AR count for each row.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ env:
     - BUILD_TYPE=Debug
     - TECA_DIR=/travis_teca_dir
     - TECA_PYTHON_VERSION=2
-    - TECA_DATA_REVISION=46
+    - TECA_DATA_REVISION=48
   matrix:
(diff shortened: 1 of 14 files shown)


taobrienlbl:
@burlen - I just tested this out.  The test I ran worked nicely and the file contains exactly what I'd expect: nice!

burlen:
> @burlen - I just tested this out. The test I ran worked nicely and the file contains exactly what I'd expect: nice!

@taobrienlbl Thank you for testing this out, glad it is doing what you need! newer  commits add support for reading the data back in, adds a regression test, and fixes some issues that popped up along the way. the regression test depends on #289 so that branch needs to be merged first.

burlen:
failing tests should be resolved by #284

burlen:
no new changes, rebasing from the rebased python_algorithm

burlen:
non new commits, rebased from master, and changed merge target to master

PR #289, opened 2019-09-27
Python algorithm

burlen:
adds a base class containing all plumbing code and satisfying teca'isms to simplify python development.


burlen:
failing tests should be resolved by #284

burlen:
OK, I'll work on merging this, but something is arwy with the history. perhaps there was a merge from master? at any rate I think I can get it straightened and merged this am.

burlen:
btw, no worries about that, but in the future please rebase as opposed to merge

taobrienlbl:
Sorry about that @burlen - I did a merge on #285 as well, so you'll probably need to check that.  I'll check with you next time on rebase vs merge: I obviously don't quite have it right yet.  I did merge b/c I thought that rebase would cause a history-rewrite on the remote branch.  I'll ask you about this tomorrow.

burlen:
> Sorry about that @burlen - I did a merge on #285 as well, so you'll probably need to check that. I'll check with you next time on rebase vs merge: I obviously don't quite have it right yet. I did merge b/c I thought that rebase would cause a history-rewrite on the remote branch. I'll ask you about this tomorrow.

Not a problem! Easy for me to resolve. It's very valuable that you actually tried it out as part of your review, and kick started a new round of tests which us saves time today.

don't worry about #285, that one is complicated because it depends on this one, not master. it will require some finagling for that reason.  I've already updated for the most recent push here.

You are correct, rebase rewrites history, which complicates things for the other parties. what you did here is generally the right thing to do when collaborating on something. Keeping topic branches free of merges is important competing interest. 

Which strategy to take is situational, and somewhat arbitrary choice. merging being the safe one. When it's easy to do, as it is here, my take is that we should favor rebase. Keeps things cleaner and makes it easier to deal with in a number of situations compared to merges. One thing we could do is to check in with the developers involved and make the call to merge or rebase when collaborating based on what ever works best in that particular situation.

taobrienlbl:
nice, and thanks @burlen for the patient explanation.  Good to know that my understanding wasn't totally off.

> One thing we could do is to check in with the developers involved and make the call to merge or rebase when collaborating based on what ever works best in that particular situation.

That makes a lot of sense. I'll make a point to do that for future PRs.

Issue #300, opened 2019-10-03
progress bar for TECA apps


taobrienlbl:
@ndkeen and @taobrienlbl were just talking about the idea of TECA apps (i.e., the bayesian AR detector) having some sort of status indicator.  The current behavior is that after the run initializes, no output is given. For longer runs, it can be unclear whether execution is occurring or whether things have stalled (e.g., filesystems hanging at NERSC).

Thoughts @burlen ?

ndkeen:
Yes, it would be nice if there was "something" written to a logfile or stdout/stderr to let us know progress. I think that's nice in general, but specifically when a "larger" run crashes, I don't have much to go on regarding what actually happened. Was it just getting started? Near the end? Crashing in same location?  It can also be nice to have timestamps written along with status report as well which could help -- for example if might indicate if the filesystem suddenly slows down during a run.

burlen:
OK. there is such a feature in place. it's not exactly a progress bar, but  gives some indication of forward progress. let me see if I can turn it on in the app.

also, running "ls -lah" in the output dir as the run progresses shows the time each file was created, and as the run progresses more files will show up. that could answer most of your questions.

if you're seeing crashes perhaps there's new bug. I'll take a look first thing in the morning.

burlen:
#301 enables progress reporting in the Bayesian ar detector.

I again verified that things are running smoothly. I would like to help with crashes you mention. It may be a run configuration issue (does running with the defaults crash?), or a bug. Either way I bet I could tell you very quickly what is going wrong. However, the only way I can help you is if you share the information needed to reproduce the issue.

PR #301, opened 2019-10-04
bayesian_ar_detect add flag to control verbose terminal output

burlen:
as requested in #300 

* adds some detail to the verbose option of the executive. which when run
  in verbose mode now describes how the index space is partitioned to
  ranks and reports the details of each request as it is accessed.

* adds --verbose to the bayesian_ar_detect command line
  this defaults to off for 2 reasons.

    1. Terminal will become trashed at large scales, limitting the
       usefulness of the output
    2. Especially with the thread pool, there is a cost (communication &
       synchronization) when generating the report


burlen:
rebased from master

PR #304, opened 2019-10-28
cf_reader read string attributes

burlen:
WIP -- need data to test this. The changes may or may not be complete at this point


burlen:
tested by Travis and included in #308

PR #308, opened 2019-11-20
ARTMIP output format

taobrienlbl:
modifications for compliance with ARTMIP: 
 * added ability to add attributes to teca_binary_segmentation variable
 * added ar_binary_tag (segments ar_probability at a user-specified
   level) and metadata to netcdf output
 * added option to turn compression on for point arrays
 * added test for cf_writer::set_compression_level()

### DIFF
--- a/alg/teca_binary_segmentation.cxx
+++ b/alg/teca_binary_segmentation.cxx
@@ -318,13 +318,13 @@ const_p_teca_dataset teca_binary_segmentation::execute(
 
     // do segmentation
     size_t n_elem = input_array->size();
-    p_teca_unsigned_char_array segmentation =
-        teca_unsigned_char_array::New(n_elem);
+    p_teca_char_array segmentation =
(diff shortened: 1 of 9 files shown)


taobrienlbl:
@burlen - I think you haven't started reviewing yet, but in case you have, FYI that I just force pushed a change to `teca_binary_segmentation.cxx`.  The new `ar_binary_tag` variable that I added to the cf_writer output needs to be of type `nc_byte` and not `nc_ubyte` in order to comply with the formatting guidelines.

taobrienlbl:
I just had to push another two mods: I added the "description" attribute to "binary_ar_tag" for `teca_bayesian_ar_detect`, and I cherry-picked the commit from #304 

Btw, #304 did the trick -- I tested it on the data that were causing the original problem, and it works!  I'll comment there momentarily.

taobrienlbl:
Thanks @burlen - just merged.

PR #312, opened 2019-12-06
fix type issues in Python bindings

burlen:


burlen:
good to merge once tests have passed.

Issue #305, opened 2019-11-07
wv_cc not found when using 1 parameter in teca_bayesian_ar_detect


taobrienlbl:
In attempting to get the teca_bayesian_ar_detect algorithm to work with only one parameter table row, I constructed a python pipeline and am using teca_dataset_source (w/ a teca_table set as its dataset).  The pipeline works fine when I run with a table with 2 rows, but it fails when running with a table with 1 row:

ERROR: [0:140340862256896] [/opt/TECA/alg/teca_bayesian_ar_detect.cxx:180 2.2.1-228-gc6142a3]
ERROR: finalize failed, proability array "ar_probability" not found

This behavior is documented in #215. This occurs because the parameter_table_reduction::operator() routine (reduce()) is not called when only 1 parameter is involved in the reduction (from ~line 263 of teca_index_reduce.cxx):

```c++
// --------------------------------------------------------------------------
const_p_teca_dataset teca_index_reduce::reduce_local(
    std::vector<const_p_teca_dataset> input_data) // pass by value is intentional
{
    unsigned long n_in = input_data.size();

    if (n_in == 0)
        return p_teca_dataset();

    if (n_in == 1)
        return input_data[0];
```

I attempted to modify the if(n_in == 1) part as follows in order to force a call to the operator() routine:

```c++
    if (n_in == 1)
            // my modification to force reduce() to be called
            TECA_PROFILE_METHOD(128, this, "reduce",
                input_data[0] = this->reduce(input_data[0], input_data[n_in-1]);
                )
        return input_data[0];
```

But this results in a new error that 'wv_cc' isn't present. It's like the request isn't working properly when the table has only one row.

The pipeline that I'm using follows:

```python

import teca
import numpy as np



class TECAComponentCount(teca.teca_python_algorithm):
    
    def __init__(self):
         return
        
    def get_execute_callback(self):
        def execute(port, data_in, req):
            # get the input mesh
            in_mesh = teca.as_teca_cartesian_mesh(data_in[0])
            
            ts = in_mesh.get_time_step()
            tsa = teca.teca_variant_array.New(np.array([ts],'int32'))
            
            info = in_mesh.get_information_arrays()
            ar_count = np.array(info.get('ar_count'))[0]
            
            ncc = teca.teca_variant_array.New(np.array([ar_count],'int32'))
            
            tab = teca.teca_table.New()
            tab.append_column('step', tsa)
            tab.append_column('count', ncc)
            
            return tab
        
        return execute
    
[...]

Issue #221, opened 2019-05-02
better document reduce strategy


burlen:
in recent mr #203 we fixed a bug where the reduce override was only called if there are 2 datasets to reduce. thus far we implemented all the reductions such that 4 cases are considered

in pseudocode
```
def reduce(left, right):
    if (left && right):
        # case 1, both valid
    else if (!right)
        # case 2, only left valid  
    else if (!left)
         # case 3, only right valid
    else
         # case 4, neither valid
```

however, in our dispatching implementation there are only 2 cases that occur,  case 1 and case 2. case 3 and 4 would potentially be error cases.

questions:
1. should we document that user's only need to implement the first 2 cases? or all 4?
2. is there ever a time, other than an error condition, when we want to allow null dataset to be returned?

PR #313, opened 2019-12-06
Streaming mode

burlen:
streaming mode parallelizes reductions and CF writer when writing multiple time steps to a single file. The threaded algorithm waits on a small number of tasks and processes results as they become available. The bugs #215, #305, #221, (X?) need to be fixed as a part of this.


burlen:
@taobrienlbl ready for stress test. would be helpful to see if you can break it.

burlen:
@taobrienlbl as I think about this a little more, before we stress test this, it would make sense to add the MPI parallel I/O part. Sorry for jumping the gun.

burlen:
@taobrienlbl  The following error occurs because the time dimension is unlimtted.
```
NetCDF: Attempt to extend dataset during NC_INDEPENDENT I/O operation. Use nc_var_par_access to set mode NC_COLLECTIVE before extending variable.
```
because we anticipate time steps to be processed in a non-deterministic order NC_COLLECTIVE mode is not ideal. I may be able to switch in between these modes but there are some warnings about switching modes impacting performance(https://www.unidata.ucar.edu/software/netcdf/docs/group__datasets.html#ga6dc46e4ab82584360518db5cb0cad841)

We know up front how many time steps will be written into each file, hence fix time dimension is possible. Also the code to read a netcdf variable does not depend on if the variable has an unlimited dimension. For that reason I think all apps will be able to read our data whether we write it with an unlimited time dimension or a fixed one. You mentioned that having an unlimited time dimension is necessary though, and I'd like to know more.

If the idea is that some codes could extend the file in place and add new time steps, the same result could be achieved by making a new file. My guess is that while making a new file requires more disk space temporarily, it has comparable performance than extending an existing one in place.

On today's call can we review this issue? what are the cases where an unlimited dimension is really needed? What are the use cases for adding time steps to an existing file wrt to teca output?

burlen:
@taobrienlbl it may be possible to concatenate files with fixed time dimension using the following approach
```
ncks -O --mk_rec_dmn time in.nc out.nc
ncrcat -h out.nc file_1.nc file_2.nc file_out_1_2.nc
```

burlen:
data for validation:

```
/global/cscratch1/sd/mwehner/machine_learning_climate_data/ /global/cscratch1/sd/mwehner/machine_learning_climate_data/All-Hist/CAM5-1-0.25degree_All-Hist_est1_v3_run1/IVT/
```

alg/teca_table_reduce.cxx
```diff
@@ -64,24 +64,21 @@ p_teca_dataset teca_table_reduce::reduce(
 
     p_teca_table output_table;
 
-    bool left = left_table && *left_table;
-    bool right = right_table && *right_table;
-
-    if (left && right)
+    if (left_table && right_table)
```
taobrienlbl, line 67: previously it looks like you were checking for the existence of the `left_table` pointer and a valid value for it; now it just looks like you're checking for the pointer existence.  If I'm understanding that right, was the previous code redundant in some way?  Or has the `teca_table` changed in some way that it has more internal safety checks?  (And this isn't something that needs to be addressed prior to merge: it's more a curiosity.)

apps/teca_cf_restripe.cpp
```diff
@@ -0,0 +1,320 @@
...
+        if (mpi_man.get_comm_rank() == 0)
+        {
+            TECA_ERROR(
+                "missing file name pattern for the NetCDF CF writer. "
+                "See --help for a list of command line options.")
+        }
+        return -1;
+    }
+
+    // if no point arrays were specified on the command line by default
```
taobrienlbl, line 209: the comment on line 209 seems to suggest that the code from lines 211-246 should do something; but that code is commented out.  I'd suggest revising the comment on 209 to reflect what the currently uncommented code is doing, and it might make sense to remove the commented code if it isn't going to be used in this release.

core/teca_threaded_algorithm.cxx
```diff
@@ -201,18 +210,49 @@ const_p_teca_dataset teca_threaded_algorithm::request_data(
...
+            int n_tasks_remaining = 0;
+            do
+            {
+               // get the requested data. will block until it's ready.
+               std::vector<const_p_teca_dataset> input_data;
+
+               n_tasks_remaining = this->internals->thread_pool->wait_some(
+                    this->stream_size, this->poll_interval, input_data);
+
+                // when stremaing recycle last round's output, this is
```
taobrienlbl, line 222: stremaing -> streaming

core/teca_threaded_algorithm.h
```diff
@@ -57,21 +57,47 @@ class teca_threaded_algorithm : public teca_algorithm
...
+    // for completed tasks
+    TECA_ALGORITHM_PROPERTY(long long, poll_interval);
+
     // explicitly set the thread pool to submit requests to
     void set_data_request_queue(const p_teca_data_request_queue &queue);
 
 protected:
     teca_threaded_algorithm();
 
+    // stremaing execute. streaming flag will be set when there is more
```
taobrienlbl, line 76: stremaing -> streaming

io/teca_cf_writer.cxx
```diff
@@ -623,53 +213,118 @@ std::vector<teca_metadata> teca_cf_writer::get_upstream_request(
...
+    {
+        TECA_ERROR("The arrays to write have not been specified")
+        return up_reqs;
+    }
 
-    // estimate the number of files
-    long n_steps_left = n_indices_up % this->steps_per_file;
-    long n_files = n_indices_up / this->steps_per_file;
+    // get the extent describing the size of the output file
+    // this can come form a few places: the request takes precedence,
```
taobrienlbl, line 225: form -> from

io/teca_cf_writer.h
```diff
@@ -10,8 +10,48 @@
...
+number of time steps - specified by the pipeline control index_initializer key
+                       found in metadata produced by the source (eg CF reader)
+
+extent - 6 64 bit integers defining the 3 spatial dimensions of each time step
+         found in metadata produced by the source (eg CF reader)
+
+point arrays - list of strings naming the point centered arrays that will be
+               written. set by the user prior to execution by writer properties.
+
+infomration arrays - list of strings naming the non-geometric arrays that will
```
taobrienlbl, line 42: infomration -> information

taobrienlbl, line 42: how are the dimensions of the information arrays defined?  It seems obvious that the point arrays would inherit these from the input data, and the information arrays need to have their dimensions defined.  It might be useful to briefly comment on what part of the API is responsible for setting the dimension names/sizes/metadata for information array dimensions.

io/teca_cf_writer.h
```diff
@@ -58,22 +98,34 @@ class teca_cf_writer : public teca_threaded_algorithm
...
-    // or equal to 0
+    // sets the compression level used for each variable compression is not
+    // used if the value is less than or equal to 0. This feature requires
+    // collective writes and is incompatible with out of order execution,
+    // and hence currently not suppoorted.
     TECA_ALGORITHM_PROPERTY(int, compression_level)
 
+    // specify the arrays to write. a data array is only written to disk if
+    // it is included in this list. it is an error to not specify at least
+    // one point cenetred array to write
```
taobrienlbl, line 114: cenetred -> centered

python/teca_py_algorithm.h
```diff
@@ -369,6 +390,90 @@ class reduce_callback
...
+
+        // convert the return
+        int i_own = 0;
+        p_teca_dataset *tmp_data = nullptr;
+        if (!SWIG_IsOK(SWIG_ConvertPtrAndOwn(ret,
+            reinterpret_cast<void**>(&tmp_data),
+            SWIGTYPE_p_std__shared_ptrT_teca_dataset_t, 0, &i_own))
+            || !tmp_data)
+        {
+            // this is not necessarilly an error
```
taobrienlbl, line 462: necessarilly -> necessarily

alg/teca_table_reduce.cxx
burlen: good question.  The reason for the change is that in the new streaming mode one of the tables can be empty, and it is a case that will now occur regularly.
```
bool right = right_table && *right_table;
```
In the statement `right_table` evaluates true if the shared pointer is not null. `*right_table` evaluates true if the table it points to is not empty. I implemented an automatic conversion to bool. Here is the code for the latter
```
    explicit operator bool() const noexcept
    { return !this->empty(); }
```
These type of conversions while convenient are one of the more esoteric features of C++,  can be tricky to read (especially when mixed with smart pointers), and have some gotchas that can cause problems if one is not careful with them(the explicit helps). For better or worse I ignored all that and implemented it. We could potentially remove such conversions from the code base to improve readability and make it harder to use incorrectly.

burlen:
> @burlen : this is a tremendous amount of work that is well written and documented! I added a few inline comments to address. In addition to these, I wonder if it is worth considering a different name for `teca_cf_time_step_mapper.cxx` -- perhaps `teca_cfwriter_time_step_mapper.cxx`, since it appears that this is utility code specifically for the writer. Part of the motivation for suggesting this is that I'm wondering if we'll eventually need something like `teca_cfreader_time_step_mapper.cxx` to support work on #353 . This looks ready to merge to me, once the inline comments are addressed. I'll suggest that we merge now, run some large scale tests on cori, and create separate issues for anything that comes up. Thanks!

@taobrienlbl  Thanks for the detailed review! Very helpful to have eyes on it. I agree with all your suggestions, and will address asap. In the meantime if you see/think of anything else, feel free to add more.

burlen:
> Looks good to me

Thanks Abdel.

alg/teca_table_reduce.cxx
taobrienlbl: ah, got it, that makes a lot of sense.  No, I don't think there's any need to change what's there.  Thanks for explaining.

burlen: Now that I'm thinking about this more, the case that comes up in streaming mode is where the right table is a nullptr. I don't recall why I took the check for the empty table out. I think it is an edge case that doesn't come up too often. However, I'll do my best to double check the code to make sure there are no lurking issues here.

burlen: I added the check for empty tables back in.

apps/teca_cf_restripe.cpp
burlen: done

core/teca_threaded_algorithm.cxx
burlen: found this misspelling in 3 other places. fixed

core/teca_threaded_algorithm.h
burlen: got it

io/teca_cf_writer.cxx
burlen: done

io/teca_cf_writer.h
burlen: done

burlen: added reference to size attribute

python/teca_py_algorithm.h
burlen: fixed a couple occurrences of this one

io/teca_cf_writer.h
burlen: done

burlen:
@taobrienlbl I merged the work, thanks for the review. I think you have a good point about the names of the supporting classes like teca_cf_time_step_mapper potentially conflicting with the reader. I left dealing with it until later. I think I will move these into a teca_cf_writer_util.h/cxx file and put them in a name space.

PR #320, opened 2020-01-23
pre release bug fixes

burlen:
fixes that came about from working on the tutorial and working with Abdel


burlen:
@taobrienlbl @elbashandy FYI. I don't think that there is anything too tricky with this merge, and I will merge it now to get the release out the door, but there is some useful Python functionality here to be aware of. Also, ctest results are now submitted to 
https://cdash.nersc.gov/index.php?project=TECA
(instead of cdash.hpcvis.com) any cdash web site issues can be addressed by sending a ticket to nersc through the usual channel. If there is some problem with cdash at nersc we can switch back by reverting the associated commit in this PR.

PR #330, opened 2020-02-11
A polymorphic implementation of python algorithm

burlen:
use getter's to wrap `report`, `request` and `execute` methods that may be overriden.

### DIFF
--- a/alg/teca_python_algorithm.py
+++ b/alg/teca_python_algorithm.py
@@ -1,17 +1,17 @@
 
 class teca_python_algorithm(object):
     """
-    The base class used for writing new algorithms in Python.
-    Contains plumbing that connects user provided callbacks
-    to an instance of teca_programmable_algorithm. Users are
-    expected to override one or more of get_report_callback,
-    get_request_callback, and/or get_execute_callback. These
-    methods return a callable with the correct signature, and
-    use a closure to access class state.
+    The base class used for writing new algorithms in Python.  Contains
+    plumbing that connects user provided overrides to an instance of
+    teca_programmable_algorithm. Users are expected to override one or more of
+    report, request, and/or execute.
     """
+
     @classmethod
     def New(derived_class):
-        """ factory method returns an instance of the derived type """
+        """
+        factory method returns an instance of the derived type
+        """
         dc = derived_class()
         dc.initialize_implementation()
         return dc
@@ -36,7 +36,9 @@ class teca_python_algorithm(object):
         self.impl.set_execute_callback(self.get_execute_callback())
 
     def __getattr__(self, name):
-        """ forward stuff to the programmable algorithm """
+        """
+        forward calls to the programmable algorithm
+        """
 
         # guard against confusing infinite recursion that
         # occurs if impl is not present. one common way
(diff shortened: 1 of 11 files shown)


burlen:
@elbashandy I think this will help with your use case, see 
https://github.com/LBL-EESA/TECA/pull/330/files#diff-bb74aa816d8f290b115ca53ef9d932b8

alg/teca_python_algorithm.py
```diff
@@ -51,58 +55,73 @@ def __getattr__(self, name):
...
+        """
+        return 1
+
+    def report(self, port, md_in):
+        """
+        return the metadata decribing the data available for consumption.
+        Override this to customize the behavior of the report phase of
+        execution. The default passes metadata on the first input through.
+        """
+        import teca_py
```
taobrienlbl: I'd suggest moving this import to the top of the file, following the PEP 8 style guide for imports: https://www.python.org/dev/peps/pep-0008/#imports

alg/teca_python_algorithm.py
```diff
@@ -51,58 +55,73 @@ def __getattr__(self, name):
...
+        return teca_py.teca_metadata(md_in[0])
+
+    def request(self, port, md_in, req_in):
+        """
+        return the request for needed data for execution. Override this to
+        customize the behavior of the request phase of execution. The default
+        passes the request on the first input port through.
+        """
+        import teca_py
```
taobrienlbl: I'd suggest moving this import to the top of the file, following the PEP 8 style guide for imports: https://www.python.org/dev/peps/pep-0008/#imports

alg/teca_python_algorithm.py
```diff
@@ -51,58 +55,73 @@ def __getattr__(self, name):
...
+        return [teca_py.teca_metadata(req_in)]
+
+    def execute(self, port, data_in, req_in):
+        """
+        return the processed data. Override this to customize the behavior of
+        the execute phase of execution. The default passes the dataset on the
+        first input port through.
+        """
+        import teca_py
```
taobrienlbl: I'd suggest moving this import to the top of the file, following the PEP 8 style guide for imports: https://www.python.org/dev/peps/pep-0008/#imports

alg/teca_tc_activity.py
```diff
@@ -51,93 +51,87 @@ def set_color_map(self, color_map):
...
-
-            # organize the data by year month etc...
-            regional_ACE = []
-            regional_PDI = []
-
-            # get unique for use as indices etc
-            uyear = sorted(set(year))
-            n_year = len(uyear)
-            ureg = sorted(set(zip(region_id, region_name, region_long_name)))
+        global plt
```
taobrienlbl: from line 61-68:  imports should be at the top of the file, following the PEP-8 standard.  This should also remove the need to define the imports as global on lines 61-63

alg/teca_tc_activity.py
```diff
@@ -51,93 +51,87 @@ def set_color_map(self, color_map):
...
+        # tweak matplotlib slightly
+        plt.rcParams['figure.max_open_warning'] = 0
+        plt.rcParams['legend.frameon'] = 1
+
+        # get the input table
+        in_table = teca_py.as_teca_table(data_in[0])
+        if in_table is None:
+            # TODO if this is part of a parallel pipeline then
+            # only rank 0 should report an error.
+            sys.stderr.write('ERROR: empty input, or not a table\n')
```
taobrienlbl: This doesn't need to be addressed in this revision, but we should consider for future code/revisions that a more pythonic behavior would be for an exception to be raised (e.g., `raise RuntimeError("empty input, or not a table")`).

alg/teca_tc_stats.py
```diff
@@ -44,500 +44,495 @@ def set_rel_axes(self, rel_axes):
...
-                annual_cat.append(category[yids])
-                annual_wind.append(wind[yids])
-                annual_press.append(press[yids])
-                annual_dur.append(duration[yids])
-                annual_len.append(length[yids])
-                annual_ACE.append(ACE[yids])
-
-                # global totals
-                tmp = [annual_count[-1]]
+        import matplotlib.pyplot as plt
```
taobrienlbl: imports at top of file - PEP 8 standard

alg/teca_tc_stats.py
```diff
@@ -44,500 +44,495 @@ def set_rel_axes(self, rel_axes):
...
+        import matplotlib.patches as plt_mp
```
taobrienlbl: imports at top of file - PEP 8 standard

alg/teca_tc_trajectory_scalars.py
```diff
@@ -58,362 +58,356 @@ def set_plot_peak_radius(self, plot_peak_radius):
...
-
-                    plt.plot(lon_i[kk], lat_i[kk], '.', linewidth=2, \
-                        color=edge_color,zorder=10)
+        #sys.stderr.write('teca_tc_trajectory_scalars::execute\n')
 
-                    plt.plot(lon_i[kk], lat_i[kk], '.', linewidth=1, \
-                        color='k', zorder=10, markersize=1)
-
-                    marks = wind_rad_i[0] > 1.0e-6
+        import matplotlib.pyplot as plt
```
taobrienlbl: lines 70-73: imports at top of file - PEP 8 standard

alg/teca_tc_trajectory_scalars.py
```diff
@@ -58,362 +58,356 @@ def set_plot_peak_radius(self, plot_peak_radius):
...
+        # tweak matplotlib slightly
+        plt.rcParams['figure.max_open_warning'] = 0
+        plt.rcParams['legend.frameon'] = 1
+
+        # get the input table
+        in_table = teca_py.as_teca_table(data_in[0])
+        if in_table is None:
+            # TODO if this is part of a parallel pipeline then
+            # only rank 0 should report an error.
+            sys.stderr.write('ERROR: empty input, or not a table\n')
```
taobrienlbl: see previous comment about raising exceptions instead of printing to stderr

alg/teca_tc_wind_radii_stats.py
```diff
@@ -48,170 +48,164 @@ def set_output_prefix(self, output_prefix):
...
-                q += 1
-            pr = pr[kk]
-
-            plt.subplot(121)
-            n,bins,pats = plt.hist(pr[np.where(pr > 0.0)], 24, \
-                facecolor='steelblue', alpha=0.95, edgecolor='black', \
+        track_table = teca_py.as_teca_table(data_in[0])
+
+        # plot stats
+        import matplotlib.pyplot as plt
```
taobrienlbl: lines 60-62: imports at top of file - PEP 8 standard

test/python/test_python_algorithm.py
```diff
@@ -0,0 +1,116 @@
+from teca import *
```
taobrienlbl: We should avoid wildcard imports, especially in example code, again following PEP 8: https://pep8.org/#imports

> Wildcard imports (from <module> import *) should be avoided, as they make it unclear which names are present in the namespace, confusing both readers and many automated tools.

alg/teca_python_algorithm.py
burlen: @taobrienlbl I dislike this too. the import was moved from the top of the file to here to solve a specific bug that manifested when running from a `python setup.py install` the log message in that commit 0cdf5cd6 explains in more detail. moving the import here solved the issue for both scenarios - running from the "normal" teca build/install and running from the setup.py one. If you agree, I will make a new issue to dig into this further. One of the confounding factors is likely the `from teca_py import *` in the top level teca.py module and elsewhere. I beleive that these import * are leading to the cycle.

alg/teca_python_algorithm.py
burlen: I know that was in there to sole a specific problem, it's been so long that I don't recall the details. I added an issue #332

alg/teca_python_algorithm.py
burlen: see #332

taobrienlbl:
@burlen - I didn't appreciate that making these style changes would potentially involve code refactoring. I'd prefer not to hold this PR up over something that's probably best as a separate issue. It probably makes sense to deal with adherence to the PEP-8 standard in a single, code-wide campaign (e.g., #332 as you suggest and/or in #331).

I'm going to change my review to 'approve'.

alg/teca_tc_trajectory_scalars.py
burlen: true. we should raise.

burlen:
@taobrienlbl I'm mergeing this, and I have started a new branch for pep8 fixes

PR #334, opened 2020-02-18
make ctest fail right away when any part fails

burlen:


burlen:
> Looks good!

@elbashandy thanks!

Issue #337, opened 2020-03-03
ghost cell fill utilities


burlen:
Abdel started "mesh padding" and "mesh layering" classes. The layering class implements periodic bc in z direction while the padding class implements a zero fill on the high side of the domain. These should probably be merged into one utility namespace and handle ghost cell fill operations.

Ghost zones are often needed, we have a few of cases of interest

1. periodic boundary in lon. 
2. reflected boundary condition
3. fixed boundary condition.
4. exchange data between cooperating ranks with neighboring sub domains.

## examples
A couple of examples from the jet stream sinuosity alg.  ghost cells are used to get correct medial axis transform near domain boundary. green line shows the domain boundary. in both images  artifacts appear at the left or right edges. 32 ghost cells ensure that artifacts are far enough from the valid domain so that a correct result is obtained.
### data copied across the periodic BC.
 ![ghgost_zones_per](https://user-images.githubusercontent.com/115053/75805913-9144a180-5d37-11ea-8a1e-43724500bb12.png)

### reflected BC.
![ghgost_zones](https://user-images.githubusercontent.com/115053/75805916-91dd3800-5d37-11ea-8f7e-b92724a57214.png)

PR #338, opened 2020-03-04
Implementing mesh padding and layering algs

elbashandy:
- Mesh padding algorithm:
      an algorithm that pads mesh arrays with zeroes based on the chosen
dimensions.

- Mesh layering algorithm:
      an algorithm that converts mesh arrays into multiple channel
arrays by layering or stacking copies of the array.


burlen:
see #337 for some background info

burlen:
will close this pr but intend to keep the branch around

PR #347, opened 2020-04-14
Build and test updates

burlen:
testing and build system updates


taobrienlbl:
Looks good to me -- thanks @burlen

PR #351, opened 2020-04-29
Supporting TECA_assets svn repo in TECA

elbashandy:
Moved deeplab models + png/jpg files to TECA_assets


burlen:
it's looking good so far. We need to add a cmake target to fetch assets during build. So for instance one could say "make TECA_assets" and this would trigger a `svn co` or `svn up`. Would also like such a target for TECA_data. Both these targets need to be configurable to be in or out of the default target.

See also:
https://cmake.org/cmake/help/latest/module/ExternalProject.html 
https://cmake.org/cmake/help/v3.17/module/FetchContent.html

elbashandy:
@burlen I set `TECA_DATA_REQ` to be our flag that we can set if we want to TECA_data to be downloaded. Do you think something like this could work?
```
set(TECA_DATA_REQ OFF CACHE BOOL "require TECA_data or not")
if (TECA_DATA_REQ)
    FetchContent_Declare(
        TECA_data
        SVN_REPOSITORY svn://svn.code.sf.net/p/teca/TECA_data
    )
    FetchContent_GetProperties(TECA_data)
    if(NOT TECA_data_POPULATED)
        FetchContent_Populate(TECA_data)
        add_subdirectory(${TECA_data_SOURCE_DIR})
    endif()

    set(TECA_DATA_ROOT ${TECA_data_SOURCE_DIR})
else()
    set(TECA_DATA_ROOT "/path/to/TECA_data"
        CACHE PATH "Path to TECA_data repository")
endif()

set(tmp OFF)
if (EXISTS "${TECA_DATA_ROOT}")
    set(tmp ON)
    message(STATUS "TECA_data -- available")
elseif (REQUIRE_TECA_DATA)
    message(FATAL_ERROR "TECA_data -- required but not found")
else()
    message(STATUS "TECA_data -- not available")
endif()
set(TECA_HAS_DATA ${tmp} CACHE BOOL "TECA_data is present")
```

test/travis_ci/install_osx.sh
```diff
@@ -33,3 +33,12 @@ do
...
 done
+
+svn co svn://svn.code.sf.net/p/teca/TECA_assets@${TECA_ASSETS_REVISION} TECA_assets &
+svn_pid=$!
+
+while [ -n "$(ps -p $svn_pid -o pid=)" ]
+do
+  echo -n "."
+  sleep 2s
+done
```
burlen, line 44: needs  newline

CMakeLists.txt
```diff
@@ -331,20 +333,60 @@ endif()
...
+            CONFIGURE_COMMAND "" BUILD_COMMAND ""
+            INSTALL_COMMAND "" LOG_DOWNLOAD 1)
+        ExternalProject_Get_property(TECA_assets SOURCE_DIR)
+        set(TECA_ASSETS_ROOT ${SOURCE_DIR})
+        message(STATUS "TECA_assets -- populating content at ${TECA_ASSETS_ROOT}")
+    endif()
+else()
+    message(STATUS "TECA_assets -- not available")
+endif()
+set(TECA_HAS_ASSETS ${tmp} CACHE BOOL "TECA_assets is present")
```
burlen, line 388: tmp is not initialized at the start of this block and as a result when not requiring teca assets what ever tmp is set to  by the code block dealing with teca data repo will be used here

test/python/CMakeLists.txt
```diff
@@ -356,28 +356,28 @@ teca_add_test(py_test_bayesian_ar_detect_mpi_threads
...
 teca_add_test(py_test_deeplabv3p_ar_detect_mpi_threads
     COMMAND ${MPIEXEC} -n ${HALF_TEST_CORES} ${PYTHON_EXECUTABLE}
-        ${CMAKE_CURRENT_SOURCE_DIR}/test_deeplabv3p_ar_detect.py
-        "${TECA_DATA_ROOT}/cascade_deeplab_IVT.pt"
-        "${TECA_DATA_ROOT}/ARTMIP_MERRA_2D_2017-05.*\.nc$"
-        "${TECA_DATA_ROOT}/test_deeplabv3p_ar_detect" IVT -1
+    ${CMAKE_CURRENT_SOURCE_DIR}/test_deeplabv3p_ar_detect.py
+    "${TECA_ASSETS_ROOT}/cascade_deeplab_IVT.pt"
+    "${TECA_DATA_ROOT}/ARTMIP_MERRA_2D_2017-05.*\.nc$"
+    "${TECA_DATA_ROOT}/test_deeplabv3p_ar_detect" IVT -1
```
burlen, line 378: should be indented as before

CMakeLists.txt
```diff
@@ -331,20 +333,60 @@ endif()
...
+        endif()
+    else()
+        set(tmp ON)
+        ExternalProject_Add(TECA_assets
+            SVN_REPOSITORY svn://svn.code.sf.net/p/teca/TECA_assets
+            UPDATE_COMMAND svn update
+            CONFIGURE_COMMAND "" BUILD_COMMAND ""
+            INSTALL_COMMAND "" LOG_DOWNLOAD 1)
+        ExternalProject_Get_property(TECA_assets SOURCE_DIR)
+        set(TECA_ASSETS_ROOT ${SOURCE_DIR})
```
burlen, line 382: I believe that this will make a new variable and will not update the cached variable of the same name declared above. The cmake docs confirm this. This code should be restructured so the second variable is not needed. https://cmake.org/cmake/help/v3.18/manual/cmake-language.7.html#cmake-language-variables

test/apps/test_deeplabv3p_ar_detect_app.sh
```diff
@@ -2,32 +2,33 @@
 
 if [[ $# < 2 ]]
 then
-    echo "usage: test_deeplabv3p_ar_detect_app.sh [app prefix] " \
-         "[data root] [mpi exec] [test cores]"
+    echo "usage: test_deeplabv3p_ar_detect_app.sh [app prefix] "  \
+         "[data root] [assets root] [mpi exec] [test cores]"
```
burlen, line 6: we should no longer have to pass this on the command line , use the functions you added in teca_py_config.i

doc/rtd/applications.rst
```diff
@@ -484,7 +484,7 @@ Example
 .. code-block:: bash
 
     mpiexec -np 10 ./bin/teca_tc_trajectory_scalars   \
-        --texture ../../TECA_data/earthmap4k.png      \
+        --texture ../../TECA_assets/earthmap4k.png      \
```
burlen, line 487: we no longer need to pass on the command line, the default value can be obtained using the new api you defined in teca_py_config.i

CMakeLists.txt
```diff
@@ -331,20 +333,60 @@ endif()
...
+        set(TECA_DATA_ROOT ${SOURCE_DIR})
+        message(STATUS "TECA_data -- populating content at ${TECA_DATA_ROOT}")
+    endif()
 else()
     message(STATUS "TECA_data -- not available")
 endif()
 set(TECA_HAS_DATA ${tmp} CACHE BOOL "TECA_data is present")
 
+# configure for TECA_assets
+set(TECA_ASSETS_ROOT CACHE PATH "Path to TECA_assets repository")
```
burlen, line 364: it is unconventional not to specify a value when setting a cache variable. It would be better to restructure such that this can be avoided.

CMakeLists.txt
```diff
@@ -331,20 +333,60 @@ endif()
...
+        if (EXISTS "${TECA_DATA_ROOT}")
+            set(tmp ON)
+            message(STATUS "TECA_data -- available")
+        else()
+            message(FATAL_ERROR "TECA_data -- required but user-set path not found")
+        endif()
+    else()
+        set(tmp ON)
+        ExternalProject_Add(TECA_data
+            SVN_REPOSITORY svn://svn.code.sf.net/p/teca/TECA_data
```
burlen, line 350: we need to specify the revision using a cache variable so that released code download the correct version of the data and assets.

CMakeLists.txt
```diff
@@ -331,20 +333,60 @@ endif()
...
+            UPDATE_COMMAND svn update
+            CONFIGURE_COMMAND "" BUILD_COMMAND ""
+            INSTALL_COMMAND "" LOG_DOWNLOAD 1)
```
burlen, line 353: how verbose is this? is it disruptive to display in the terminal? if it is not too bad doing so may help a user understand what's going on. I'll leave that call up to you.

doc/teca_users_guide.tex
```diff
@@ -710,7 +712,7 @@ \subsection{Command Line Arguments}
 \subsection{Example}
 \begin{minted}{bash}
 mpiexec -np 10 ./bin/teca_tc_trajectory_scalars   \
-    --texture ../../TECA_data/earthmap4k.png      \
+    --texture ../../TECA_assets/earthmap4k.png      \
```
burlen, line 715: use the new api in teca_py_config.i to fetch a default internally

CMakeLists.txt
```diff
@@ -331,20 +333,60 @@ endif()
...
-set(tmp OFF)
-set(TECA_DATA_ROOT "/path/to/TECA_data"
-    CACHE PATH "Path to TECA_data repository")
-if (EXISTS "${TECA_DATA_ROOT}")
-    set(tmp ON)
-    message(STATUS "TECA_data -- available")
-elseif (REQUIRE_TECA_DATA)
-    message(FATAL_ERROR "TECA_data -- required but not found")
+# configure for TECA_data
+set(TECA_DATA_ROOT CACHE PATH "Path to TECA_data repository")
```
burlen, line 337: should specify a default value for the cache variable

CMakeLists.txt
```diff
@@ -331,20 +333,60 @@ endif()
...
+        endif()
+    else()
+        set(tmp ON)
+        ExternalProject_Add(TECA_data
+            SVN_REPOSITORY svn://svn.code.sf.net/p/teca/TECA_data
+            UPDATE_COMMAND svn update
+            CONFIGURE_COMMAND "" BUILD_COMMAND ""
+            INSTALL_COMMAND "" LOG_DOWNLOAD 1)
+        ExternalProject_Get_property(TECA_data SOURCE_DIR)
+        set(TECA_DATA_ROOT ${SOURCE_DIR})
```
burlen, line 355: restructure this as explained in section re: assets below

burlen (changes requested):
Hi Abdel, solid progress was made, but I flagged a few issues. In addition to comments in the code, also need to install the assets by make install.

CMakeLists.txt
elbashandy: Yes it is a little disruptive especially to the progress percentage-values that we get at the building process.

doc/teca_users_guide.tex
elbashandy: I don't think we should do that because this is given as a command line argument to the `--texture` option in the `teca_tc_trajectory_scalars` app. We can modify it to `/path/to/background_image.png` if that makes sense. Please let me know if I am missing something.

doc/rtd/applications.rst
elbashandy: I don't think we should do that because this is given as a command line argument to the `--texture` option in the `teca_tc_trajectory_scalars` app. We can modify it to `/path/to/background_image.png` if that makes sense. Please let me know if I am missing something.

test/apps/test_deeplabv3p_ar_detect_app.sh
elbashandy: I applied this change only to the `test_deeplabv3p_ar_detect.py` test in [8a8bdb](https://github.com/LBL-EESA/TECA/pull/351/commits/8a8bdb2e85a86c212e899c369ed75a7750bae1a6). However I didn't do this to the app because the user is expected to pass the deeplab model path to the command line `--pytorch_deeplab_model`. That's why `[assets root]` is being passed to this bash file

burlen: You need to update the command line application. You could leave the command line option, so that one could pass the model there. The default should be that the model is taken from the assets directory. This simplifies the use of the app. Having the model available internally(in the install) is the point of this work. There's no reason not to take advantage of it to make the tools more robust and easier to use.

doc/rtd/applications.rst
burlen: You might be missing that the point of deploying the assets with TECA is that we no longer need to pass these on the command line. It's fine to leave the command line options but make them optional, with the default value pointing to the assets location. Same deal as with the pytorch model.

doc/teca_users_guide.tex
burlen: see comments above.

PR #354, opened 2020-05-07
Added initial draft of teca_vertical_integral.  

taobrienlbl:
The vertical integraion routine itself has been tested external to TECA, but the teca_vertical_integral.cxx code is still being written.


taobrienlbl:
@burlen -- this is for discussion tomorrow.  I've got code for doing vertical integrals, and I'm trying to create a TECA algorithm from it.  You'll see quite a few TODOs down at the bottom of `alg/teca_vertical_integral.cxx` indicating a number of implementation questions that I'll likely need some guidance on.

PR #359, opened 2020-05-14
Added ar_binary_tag to the deeplab algorithm

elbashandy:


burlen:
@taobrienlb

> . @burlen: how do you want to handle coordinating with your PR work on #313 ?

yup, I'll take that on, because this looks ready to go now, and I have a few edits to make on mine

PR #360, opened 2020-05-14
Removed resnet model dependence

elbashandy:
The deeplab model already represents the whole neural network architecture parameters

### DIFF
--- a/alg/teca_deeplabv3p_ar_detect.py
+++ b/alg/teca_deeplabv3p_ar_detect.py
@@ -50,8 +50,7 @@ class Bottleneck(nn.Module):
 
 
 class ResNet(nn.Module):
-    def __init__(self, nInputChannels, block, layers, os=16,
-                 pretrained=False, state_dict_resnet=None):
+    def __init__(self, nInputChannels, block, layers, os=16):
         self.inplanes = 64
         super(ResNet, self).__init__()
         if os == 16:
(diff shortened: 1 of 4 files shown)


burlen:
does this change anything in relation to binary files needed to be deployed?

elbashandy:
> does this change anything in relation to binary files needed to be deployed?

It shouldn't but because the resnet model wasn't being used at all as it's parameter values were overwritten by the deeplab model values

burlen:
> > does this change anything in relation to binary files needed to be deployed?
> 
> It shouldn't but because the resnet model wasn't being used at all as it's parameter values were overwritten by the deeplab model values
OK, so then if I understand you, the resnet model was never committed into the TECA_data nor to the  TECA_assets repo?

elbashandy:
> > > does this change anything in relation to binary files needed to be deployed?
> > 
> > 
> > It shouldn't but because the resnet model wasn't being used at all as it's parameter values were overwritten by the deeplab model values
> 
> OK, so then if I understand you, the resnet model was never committed into the TECA_data nor to the TECA_assets repo?

Oh no I need to delete those haha. I thought you were talking about the regression tests binary files

burlen:
@elbashandy please remove the uneeded files from the svn repos and don't forget to update the teca_data svn repo test version so that we don't have to pull them down when we test.

Issue #197, opened 2019-04-04
regression tests for applications


burlen:
make regression tests that exercise the applications in the apps dir. for now it is enough that these run without error. leave to the future a way to validate the output.

```
teca_convert_table.py
teca_dataset_metadata.py
teca_event_filter.py
teca_metadata_probe.cpp
teca_tc_detect.cpp
teca_tc_stats.py
teca_tc_storm_size.cpp
teca_tc_trajectory.cpp
teca_tc_trajectory_scalars.py
teca_tc_wind_radii.cpp
teca_tc_wind_radii_stats.py
```

taobrienlbl:
Let's focus on just getting a test for the probabilistic AR detector for this milestone.

burlen:
@elbashandy @taobrienlbl this issue is a very high priority. Changes in the cf writer breaking applications but there are no tests to determine what's broken. If you guys agree I'd like this issue to take precedence over other pending tasks on Abdel's queue.

burlen:
@elbashandy another option is that I take this on. I put a meeting on the calendar this morning to figure it out.

burlen:
@taobrienlbl would you please point me to 30 days of data containing ars?  I want to use this as the basis for a test dataset for the ar app regression tests. I poked around on m1517 project dir but some of the data I found had broken symlinks.

burlen:
these tests don't need baselines, but we want to ensure that they run without error on some representative data

elbashandy:
@burlen /global/cscratch1/sd/taobrien/merra_2d

taobrienlbl:
@burlen - I would use these --`/global/cscratch1/sd/mwehner/machine_learning_climate_data/All-Hist/CAM5-1-0.25degree_All-Hist_est1_v3_run1/IVT/`.  I've run TECA BARD and the TECA pytorch algorithm successfully on both, and both produce reasonable results.

@elbashandy - the files you are referring to probably got purged.  I can redownload if you need.

burlen:
see also #256

elbashandy:
@taobrienlbl I am looking into this issue, do you have ready commands for each application (or the ones that you regularly use)? I guess we can talk about having pretty small datasets to run the apps on.

burlen:
@elbashandy please see the user's guide for documentation of how to run the command line apps.. The Bayesian ar detector may not yet be documented there yet,  but should run on the same command that you run your ar detector.

burlen:
[...]

PR #366, opened 2020-06-04
Tc app bug fixes

burlen:
fixes a handful of bugs in command line apps


taobrienlbl:
and n/m about the issue, I see #197 now (was going through e-mail in reverse chronological order): thanks!

PR #372, opened 2020-06-12
Implementing regression tests for TECA apps

elbashandy:
1. Run app first
2. Run app output verifier.


elbashandy:
@burlen Please give a look at the new code and verify if this is a liable design for our future app regression tests.

burlen:
what is this about?
```
FutureWarning: pickle support for Storage will be removed in 1.5. Use `torch.save` instead
```

burlen:
this one should be fixed
```
WARNING: Metadata is missing time axis and or calendaring info. The file id will be used in file name instead.
```

elbashandy:
> what is this about?
> 
> ```
> FutureWarning: pickle support for Storage will be removed in 1.5. Use `torch.save` instead
> ```

Our code doesn't use pickle. This is related to the pytorch version that get installed using `pip3`

@burlen I left `eval` in the bash file `run_and_verify.sh` as I was facing issues with cmake, it was hard passing space-separated arguments like `--t_units "days since ..."` to the bash file. I tried adding single quotes to the space-separated string, but that didn't work as the words were broken into space separated command arguments:
```
--t_units "'days" "since" ...
```

To show the difference between  `eval` and backticks, I added `bash -vx` in the cmake `COMMAND` option. This was output for:
1. backticks:
```
91: `${@:3:$1}`
91: ++ /app/TECA_build/bin/teca_deeplabv3p_ar_detect
--pytorch_deeplab_model  /project/TECA_data/cascade_deeplab_IVT.pt
--input_regex '/project/TECA_data/ARTMIP_MERRA_2D.*.nc$'
--filename_time_template ARTMIP_MERRA_2D_%Y%m%d_%H.nc
--t_axis_variable ''\'''\'''
--t_units ''\''days' since 1979-01-01 '00:00:00'\'''
--output_file /project/TECA_data/test_deeplabv3p_ar_detect_app_output.nc
91: usage: teca_deeplabv3p_ar_detect [-h] [--input_file INPUT_FILE]
91:                                  [--input_regex INPUT_REGEX]
91:                                  [--output_file OUTPUT_FILE] [--ivt IVT]
91:                                  [--num_threads NUM_THREADS]
91:                                  [--binary_ar_threshold BINARY_AR_THRESHOLD]
91:                                  [--pytorch_deeplab_model PYTORCH_DEEPLAB_MODEL]
91:                                  [--t_axis_variable T_AXIS_VARIABLE]
91:                                  [--t_calendar T_CALENDAR] [--t_units T_UNITS]
91:                                  [--filename_time_template FILENAME_TIME_TEMPLATE]
91:                                  [--compression_level COMPRESSION_LEVEL]
91:                                  [--date_format DATE_FORMAT]
91:                                  [--first_step FIRST_STEP]
91:                                  [--last_step LAST_STEP]
91:                                  [--steps_per_file STEPS_PER_FILE]
91:                                  [--start_date START_DATE]
91:                                  [--end_date END_DATE]
91: teca_deeplabv3p_ar_detect: error: unrecognized arguments: since 1979-01-01 00:00:00'
```
2. `eval`:
```
91: eval ${@:3:$1}
91: + eval /app/TECA_build/bin/teca_deeplabv3p_ar_detect
--pytorch_deeplab_model /project/TECA_data/cascade_deeplab_IVT.pt
--input_regex '/project/TECA_data/ARTMIP_MERRA_2D.*.nc$'
--filename_time_template ARTMIP_MERRA_2D_%Y%m%d_%H.nc
--t_axis_variable ''\'''\'''
--t_units ''\''days' since 1979-01-01 '00:00:00'\'''
--output_file /project/TECA_data/test_deeplabv3p_ar_detect_app_output.nc
91: /app/TECA_build/bin/teca_deeplabv3p_ar_detect
--pytorch_deeplab_model /project/TECA_data/cascade_deeplab_IVT.pt
--input_regex /project/TECA_data/ARTMIP_MERRA_2D.*.nc$
--filename_time_template ARTMIP_MERRA_2D_%Y%m%d_%H.nc
--t_axis_variable ''
--t_units 'days since 1979-01-01 00:00:00'
--output_file /project/TECA_data/test_deeplabv3p_ar_detect_app_output.nc
91: ++ /app/TECA_build/bin/teca_deeplabv3p_ar_detect
--pytorch_deeplab_model /project/TECA_data/cascade_deeplab_IVT.pt
--input_regex '/project/TECA_data/ARTMIP_MERRA_2D.*.nc$'
--filename_time_template ARTMIP_MERRA_2D_%Y%m%d_%H.nc
--t_axis_variable ''
--t_units 'days since 1979-01-01 00:00:00'
--output_file /project/TECA_data/test_deeplabv3p_ar_detect_app_output.nc
91: STATUS: [0:140061102892352] [/project/TECA/io/teca_cf_reader.cxx:406 3.0.0-96-gfb82702]
91: STATUS: Found metadata cache "/project/TECA_data/c060b58adfe471af5e93a7083c5d7606d429abe2.tmd"
```

burlen:
@elbashandy the point was to remove eval as it's not needed. There as some miscommunication because back ticks should not be used either.

burlen:
@elbashandy here is a complete example that you could copy from that I wrote for another project.

cmake
```
  senseiAddTest(testADIOS1Flexpath
      COMMAND ${CMAKE_CURRENT_SOURCE_DIR}/testADIOS1.sh
      ${MPIEXEC} ${MPIEXEC_NUMPROC_FLAG} ${TEST_NP}
      ${PYTHON_EXECUTABLE} ${CMAKE_CURRENT_SOURCE_DIR}
      testADIOS1Flexpath.bp FLEXPATH FLEXPATH 3
    FEATURES ${ENABLE_PYTHON} ${ENABLE_ADIOS1})

  senseiAddTest(testADIOS1MPIBP
    COMMAND ${CMAKE_CURRENT_SOURCE_DIR}/testADIOS1.sh
      ${MPIEXEC} ${MPIEXEC_NUMPROC_FLAG} ${TEST_NP}
      ${PYTHON_EXECUTABLE} ${CMAKE_CURRENT_SOURCE_DIR}
      testADIOS1MPIBP.bp MPI BP 3
    FEATURES ${ENABLE_PYTHON} ${ENABLE_ADIOS1})
```

shell script
```
#!/usr/bin/env bash

if [[ $# < 8 ]]
then
  echo "test_adios.sh [mpiexec] [npflag] [nproc] [py exec] [src dir] [file] [write method] [read method] [nits]"
  exit 1
fi

mpiexec=`basename $1`
npflag=$2
nproc=$3
nproc_write=$nproc
let nproc_read=$nproc-1
nproc_read=$(( nproc_read < 1 ? 1 : nproc_read ))
pyexec=$4
srcdir=$5
file=$6
writeMethod=$7
readMethod=$8
nits=$9
delay=1
maxDelay=30

trap 'eval echo $BASH_COMMAND' DEBUG

rm -f ${file}

echo "testing ${writeMethod} -> ${readMethod}"
echo "M=${nproc_write} x N=${nproc_read}"

${mpiexec} ${npflag} ${nproc_write} ${pyexec} ${srcdir}/testADIOS1Write.py ${file} ${writeMethod} ${nits} &
writePid=$!

if [[ "${readMethod}" == "BP" ]]
then
  # with BP wait for the write side to completely finish before starting the read side
  echo "waiting for writer(${writePid}) to complete"
  wait ${writePid}
elif [[ "${readMethod}" == "FLEXPATH" ]]
then
  # with FLEXPATH wait for the writer to start before starting the read side
  # look for the files that it spits out with connection info as eviddence that
  # it is ready
  echo "waiting for writer to start ${delay}"
  while [[ True ]]
  do
    if [[ -e "${file}_writer_info.txt" ]]
    then
      break
    elif [[ ${maxDelay} -le 0 ]]
    then
      echo "ERROR: max delay exceded"
      exit -1
    else
      sleep ${delay}s
      let maxDelay=${maxDelay}-${delay}
    fi
  done
fi

${mpiexec} ${npflag} ${nproc_read} ${pyexec} ${srcdir}/testADIOS1Read.py ${file} ${readMethod}
exit 0
```

elbashandy:
@burlen @taobrienlbl I did add tests for the apps described in the documentation. Please review the code when possible.

alg/teca_tc_wind_radii_stats.py
```diff
@@ -158,7 +158,7 @@ def execute(self, port, data_in, req):
             month[-1],day[-1],year[-1]), fontweight='bold', fontsize=12)
         plt.subplots_adjust(hspace=0.35, wspace=0.35, top=0.90)
 
-        plt.savefig(self.output_prefix + 'wind_radii_stats.png')
+        plt.savefig(self.output_prefix + '_wind_radii_stats.png')
```
burlen, line 161: please do not make this change. having no separator was intentional. for instance if the user specifies '/home/abdel/' as the prefix this change would result in files named with a leading `_`. applies to other scripts where `_` was added.

test/apps/test_bayesian_ar_detect_app.sh
```diff
@@ -0,0 +1,40 @@
...
+         "[t_axis_var_option] [t_axis_var]" \
+         "[t_units_option] [t_axis_var]" \
+         "[t_units_option] [t_units]" \
+         "[output_file_option] [output_file]" \
+         "[verifier_exec] [output_file_type] [output_ref]"
+    exit 1
+fi
+
+# First command
+app_exec=${1}
```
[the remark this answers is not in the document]
elbashandy: I think they are necessary for the space separated arguments in cmake. I only used them in TECA-BARD and Deeplabv3p.. I tried multiple ways to run this and only `eval` was able to run the command successfully.

test/apps/test_bayesian_ar_detect_app.sh
```diff
@@ -0,0 +1,40 @@
...
+t_units="${9}"
+output_file_option=${10}
+output_file="${11}"
+
+# Second command
+verifier_exec=${12}
+output_file_type=${13}
+output_ref="${14}"
+
+eval ${app_exec} ${input_regex_option} ${input_regex} \
```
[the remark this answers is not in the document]
elbashandy: I think they are necessary with space-separated arguments (like --t_units) in cmake. I only used it for TECA-BARD & Deeplabv3p apps. I tried multiple ways to run the commands and only `eval` was able to run the command successfully.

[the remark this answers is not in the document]
burlen: what you are describing is a quoting issue.

test/apps/app_output_verifier.cpp
```diff
@@ -0,0 +1,137 @@
...
+    unsigned int out_file_type = atoi(argv[1]);
+    std::string app_out_file = argv[2];
+    std::string ref_file = argv[3];
+    std::string column_name;
+    if (argc == 5) column_name = argv[4];
+    
+
+    if (!teca_file_util::file_exists(app_out_file.c_str()))
+    {
+        cerr << "App Verifier Error: app output file doesn't exist" << endl;
```
burlen, line 51: please use TECA_ERROR for error messages. applies here and elsewhere cerr is used to print error messages

test/apps/test_bayesian_ar_detect_app.sh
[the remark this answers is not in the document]
elbashandy: What you are saying is doable to for the app name itself but a path to the exec file is needed to be provided by cmake, like `${CMAKE_BINARY_DIR}/${BIN_PREFIX}/` for example. Do you think we should change `app_exec` to `app_prefix + app_name`? where `app_name` is going to be hard-coded.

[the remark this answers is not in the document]
burlen: ah, good point. yes that sounds good, pass in the path.

elbashandy:
@burlen I did the modifications as requested. Please let me know if I am missing something.

burlen:
@elbashandy I already see one thing missing that I asked for - putting bug fixes into individual commits. I suggest making a list and checking each item off as you go. that way nothing gets missed.

elbashandy:
@burlen cdash is down, that's why the tests are failing. 

Regarding the code, is there something else I didn't address?

test/apps/CMakeLists.txt
```diff
@@ -0,0 +1,112 @@
...
+    LIBS teca_io teca_alg
+    COMMAND ${CMAKE_CURRENT_SOURCE_DIR}/test_tc_detect_app.sh
+    ${CMAKE_BINARY_DIR}/${BIN_PREFIX}
+    --input_regex "${TECA_DATA_ROOT}/test_tc_candidates_1990_07_0[12]\\.nc"
+    --candidate_file "${TECA_DATA_ROOT}/test_tc_candidates_app_output.bin"
+    --track_file "${TECA_DATA_ROOT}/test_tc_track_app_output.bin"
+    --lowest_lat -20 --highest_lat 20
+    "${TECA_DATA_ROOT}/test_tc_candidates_app_ref.bin"
+    "${TECA_DATA_ROOT}/test_tc_track_app_ref.bin" "storm_id"
+    REQ_TECA_DATA)
```
burlen, line 46: this requires netcdf

elbashandy: Do you mean as an input? It is set on line 40 `--input_regex "${TECA_DATA_ROOT}/test_tc_candidates_1990_07_0[12]\\.nc"`

The .bin files mentioned on 44 & 45 are reference files to be diffed with app output

test/apps/CMakeLists.txt
```diff
@@ -0,0 +1,112 @@
+
+set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin/test)
+
+include_directories(${CMAKE_CURRENT_SOURCE_DIR}
+    $<TARGET_PROPERTY:teca_alg,INTERFACE_INCLUDE_DIRECTORIES>
+    $<TARGET_PROPERTY:teca_io,INTERFACE_INCLUDE_DIRECTORIES>
+    )
+
+teca_add_test(test_deeplabv3p_ar_detect_app
+    LIBS teca_io teca_alg
```
burlen, line 10: `LIBS` aren't needed since these tests aren't compiled (same comment applies where LIBS is passed and the test isn't compiled)

test/apps/CMakeLists.txt
burlen: this test can't run without netcdf, it needs `FEATURES ${TECA_HAS_NETCDF}`

elbashandy: oh haha my bad

test/apps/CMakeLists.txt
elbashandy: Great point

test/apps/CMakeLists.txt
```diff
@@ -0,0 +1,112 @@
...
+    COMMAND ${CMAKE_CURRENT_SOURCE_DIR}/test_deeplabv3p_ar_detect_app.sh
+    ${CMAKE_BINARY_DIR}/${BIN_PREFIX}
+    --pytorch_deeplab_model ${TECA_DATA_ROOT}/cascade_deeplab_IVT.pt
```
burlen, line 13: option names don't change & there is no need to pass the option name as an argument. for example here you are passing `--pytorch_deeplab_model ${TECA_DATA_ROOT}/cascade_deeplab_IVT.pt` instead you can just pass `${TECA_DATA_ROOT}/cascade_deeplab_IVT.pt` (please change this here and elsewhere where you have done that).

elbashandy: Don't you think it's cleaner that way? where any developer can tell what arguments were passed to the app in the cmake file without having to look into the bash file and try to connect the options and values together

burlen: no, I would not have asked you to change it if I thought that. it makes everything longer than it needs to be. the cmake script and you shell scripts will be significantly shorter and easier to read. it's something that I would be tempted to let slide but since you have more work to do, we might as well clean it up.

burlen: btw, I have similar thoughts about ${app_exec} in your shell script. Using a variable here adds an unnecessary line of code, and makes the necessary line of code (the one that runs the app) less clear.

elbashandy: makes sense

apps/teca_cartesian_mesh_diff.cpp
```diff
@@ -0,0 +1,95 @@
...
+                << "Usage:" << endl
+                << "teca_cartesian_mesh_diff [test file type] [test file] "
+                   "[reference file]" << endl
+                 << endl;
+        }
+        return -1;
+    }
+
+    // parse command line
+    unsigned int t_file_type = atoi(argv[1]);
```
burlen, line 40: rather than having a code to name file type use `teca_file_util::extension` to get the file extension directly from the test and reference files.

apps/teca_cartesian_mesh_diff.cpp
```diff
@@ -0,0 +1,95 @@
...
+        return -1;
+    }
+
+    p_teca_cf_reader cf_reader = teca_cf_reader::New();
+
+    p_teca_cartesian_mesh_reader m_reader = teca_cartesian_mesh_reader::New();
+
+    p_teca_dataset_diff diff = teca_dataset_diff::New();
+
+    p_teca_cartesian_mesh_reader ref_reader = teca_cartesian_mesh_reader::New();
```
burlen, line 62: handle the case where the reference file is in NetCDF CF2 format. This will let us test over the time dimension. as with the input files use teca_file_utils::extension to determine which reader is needed.

burlen:
@elbashandy This morning I cleaned up the bayesian ar detector test, and added an MPI parallel invocation. I also removed eval there and in the tc detctor test. See 374ab79,  972d981

Note that the use of the `bin` format currently precludes parallel processing therefor you'll need to address this (see comment in code) and use the new cropped dataset I made today (see #392).

test/apps/CMakeLists.txt
burlen: see 374ab79 , addresses these issues for the Bayesian AR detector test

test/apps/test_bayesian_ar_detect_app.sh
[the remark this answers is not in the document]
burlen: cmake_parse_arguments (used by teca_add_test) skips "". this is an unfortunate feature of cmake. one can avoid this by calling add_test directly (eg see Python connected components test). Other tests use an optional command line argument set to the empty string by default (eg see CF reader test). There are a number of other easy solutions to avoid use of eval. 374ab79 is a simple solution since it avoids the issue altogether.

burlen:
@elbashandy let's make sure all of the apps have tests and for apps that run in both serial and parallel we have both serial and parallel tests. I am making the following table to make it easier to keep track, please add X to indicate that the test is implemented.

| Application                                     | has serial test | has MPI test |
|----------------------------------------------------|---------------|-------------|
| ~~teca_ar_detect.cpp~~                      |  -               |   -           |                 
| teca_bayesian_ar_detect.cpp              |  X              |   X          |
| teca_cartesian_mesh_diff.cpp             |                   |   -           |
| teca_cf_restripe.cpp                            |                   |                |
| teca_convert_table.in                          |                   |    -           |
| ~~teca_dataset_metadata.in~~           |  -                |   -           |
| teca_deeplabv3p_ar_detect.in             |  X              |                |
| teca_event_filter.in                               |                   |   -           |
| ~~teca_metadata_probe.cpp~~           |  -                |  -            |
| ~~teca_profile_explorer.in~~                | -                 |  -            |
| teca_table_diff.cpp                               |                    |  -            |
| teca_tc_detect.cpp                               |  X                |               |
| teca_tc_stats.in                                    |   X               |  -            |
| teca_tc_trajectory.cpp                          |   X               |  -            |
| teca_tc_trajectory_scalars.in                |   X               |               |
| teca_tc_wind_radii.cpp                         |   X               |               |
| teca_tc_wind_radii_stats.in                  |   X               |  -            |
| teca_temporal_reduction.in                  |                    |                |

key : X test is done, - will not test

test/apps/CMakeLists.txt
```diff
@@ -0,0 +1,110 @@
...
+    --input_regex "${TECA_DATA_ROOT}/ARTMIP_MERRA_2D.*\.nc$"
+    --filename_time_template "ARTMIP_MERRA_2D_%Y%m%d_%H.nc"
+    --t_axis_variable ''
+    --t_units "days since 1979-01-01 00:00:00"
+    --output_file "${TECA_DATA_ROOT}/test_deeplabv3p_ar_detect_app_output.nc"
```
burlen, line 17: Please have the app write intermediate results to the testing directory(cwd) rather than TECA_DATA_ROOT

test/apps/test_deeplabv3p_ar_detect_app.sh
```diff
@@ -0,0 +1,47 @@
...
+verifier_exec="${app_prefix}/teca_cartesian_mesh_diff"
+
+${app_exec} ${model_option} ${model_file} \
+    ${input_regex_option} ${input_regex} \
+    ${f_time_template_option} ${f_time_template} \
+    ${t_axis_var_option} "${t_axis_var}" \
+    ${t_units_option} "${t_units}" \
+    ${output_file_option} ${output_file}
+
+${verifier_exec} ${output_file_type} ${output_file} ${output_ref}
```
burlen, line 47: please clean up the intermediate test file(s) after the test. this will ensure that they can't be accidentally picked up & re-used in mpi test

elbashandy:
@burlen I think I covered everything that you requested. Please let me know if I missed something.

The only thing I see left is to implement the cases in `teca_cartesian_mesh_diff` app tests where I need to pass different formats `.nc & .bin` and compare them together. I tried adding these tests but they were passing even with the tests that should fail. This needs further investigation.

burlen:
@elbashandy before going further, please fill in the table with the tests that you finished.

elbashandy:
> @elbashandy before going further, please fill in the table with the tests that you finished.

@burlen done

burlen:
please add an MPI test for Cartesian mesh diff. what is meant by ! in the table, which cases are missing?

elbashandy:
@burlen I added a test for cartesian_mesh_diff but only compares `.bin` files.

I mentioned this here

> @burlen I think I covered everything that you requested. Please let me know if I missed something.
> 
> The only thing I see left is to implement the cases in `teca_cartesian_mesh_diff` app tests where I need to pass different formats `.nc & .bin` and compare them together. I tried adding these tests but they were passing even with the tests that should fail. This needs further investigation.

burlen:
In the latest runs
```
The following tests FAILED:
	 94 - test_bayesian_ar_detect_app_mpi (Failed)
	 95 - test_tc_detect_app_serial (Timeout)
	 96 - test_tc_detect_app_mpi (Failed)
	106 - test_temporal_reduction_app_monthly_average_mpi_thread (Failed)
	108 - test_temporal_reduction_app_monthly_minimum_mpi_thread (Failed)
	110 - test_temporal_reduction_app_monthly_maximum_mpi_thread (Failed)
	160 - py_test_deeplabv3p_ar_detect (Failed)
	161 - py_test_deeplabv3p_ar_detect_mpi (Failed)
Submit files (using https)
```
Please make sure that the tests are working on all dashboards. I just updated the baseline for the Bayesian AR detector. However, here the serial Bayseain AR tests pass but not the parallel one. This is because Ubuntu 18.04 image is missing MPI enabled netcdf. Please fix 18.04 image or transition to 20.04. Re: the other tests, please look into and address failures.

burlen:
> @burlen I added a test for cartesian_mesh_diff but only compares .bin files.
>  ... where I need to pass different formats .nc & .bin and compare them together.

We don't need to do this. The goal is to test the diff app in 2 ways.

1. when given the same data set it should pass
2. when given different data sets it should fail.

run the tests in serial and MPI parallel. please update the table when the parallel test is complete.

test/python/test_temporal_reduction.py
```diff
@@ -16,7 +16,7 @@
 argc = len(sys.argv)
 if argc < 8:
     sys.stderr.write('usage: app [in file regex] [out file base] '
-                     '[n threads] [steps per file] [interval] [operator] '
+                     '[steps per file] [n threads] [interval] [operator] '
```
burlen, line 19: why is this needed?

elbashandy: They are switched in the code where `steps_per_file = int(sys.argv[3])` & `n_threads = int(sys.argv[4])`

burlen: oops, I see it now.

test/apps/test_deeplabv3p_ar_detect_app.sh
```diff
@@ -0,0 +1,36 @@
...
+
+app_prefix=${1}
+data_root=${2}
+
+if [[ $# -eq 4 ]]
+then
+    mpi_exec=${3}
+    test_cores=${4}
+    launcher="${mpi_exec} -n ${test_cores}"
+    steps_per_file="--steps_per_file 4"
```
burlen, line 18: Do we need a different number of steps per file in the MPI case? Unless there's a good reason to do this, it would be better to run the tests in the same way.

elbashandy: Oh yea I did it before updating to Parallel-NetCDF. Will remove it

elbashandy:
> I saw a few things, but ran out of time today to finish the review. Let's get the tests passing, mpi tests implemented, and I'll make another review then.

@burlen according to the table, all tests are implemented including the mpi ones except `teca_cf_restripe`. Is there another mpi test that should be implemented?

elbashandy:
@burlen The `teca_tc_detect` app regression test is timing out. I guess the reason is that it does two step processing: Getting a) tc candidates -> b) tc tracks

Should we increase the `CTEST_TEST_TIMEOUT`?

burlen:
> Should we increase the CTEST_TEST_TIMEOUT?

Your new test should take the same time as the existing one. why is it taking longer?

elbashandy:
> > Should we increase the CTEST_TEST_TIMEOUT?
> 
> Your new test should take the same time as the existing one. why is it taking longer?

My guess is that it does an extra step which is getting the tracks after finding the candidates. Let me know if I am missing something.

burlen:
> My guess is that it does an extra step which is getting the tracks...

Guessing about the causes of issues is not acceptable approach. We can easily know the answer with 100% certainty.
 
My experience is that the trajectory stage completes in seconds when constructing the tracks for decades of storms. The existing trajectory stage regression tests - including your new app test - takes only milliseconds to complete. In your new test the time will be dominated by the candidate stage, and hence the expected time would be perhaps slightly longer than the existing candidate stage test assuming that you use the same inputs.

Why is your test taking so much longer than the existing candidates test?

elbashandy:
> erience is that the trajectory stage completes in seconds when constructing the tracks for decades of storms. The existing trajectory stage regression tests - including your new app test - takes o

You are absolutely right. My bad, I shouldn't have resolved to increasing the timeout before making sure that it is operating correctly in the first place.

Will investigate this further.

elbashandy:
I found out the reason that the tc_detect app was taking too long was because app test was not setting the start_index & end_index for map_reduce. Setting these values sped things up.

@burlen I don't think I can use the same baseline of teca_tc_candidates regression_test as there's a difference in the pipeline structure where the candidate writer in the app doesn't go through a teca_table_calendar component as the regression test one. Comparing the app output with the existing baseline gives an error about missing calendar columns:

```
ERROR: [0:140256193426624] [/project/TECA/alg/teca_dataset_diff.cxx:278 3.0.0-142-gd44b74a]
95: ERROR: The baseline table has 17 columns while test table has 12 columns. Columns "year", "month", "day", "hour", "minute" are missing
95: ERROR: [0:140256193426624] [/project/TECA/alg/teca_dataset_diff.cxx:239 3.0.0-142-gd44b74a]
95: ERROR: Failed to compare tables.
```

I will go ahead and add a new baseline that reflects the app output.

burlen:
there seems to be a prpblem with the deeplab  mpi test, it takes longer than the serial test
```
      Start  91: test_deeplabv3p_ar_detect_app_serial
 1/25 Test  #91: test_deeplabv3p_ar_detect_app_serial .....................   Passed   21.56 sec
      Start  92: test_deeplabv3p_ar_detect_app_mpi
 2/25 Test  #92: test_deeplabv3p_ar_detect_app_mpi ........................   Passed  124.89 sec

```

elbashandy:
@burlen great catch! Will look into it

elbashandy:
@burlen Is this on your local machine? It's interesting it doesn't show on travis-ci fedora test:

```
        Start  91: test_deeplabv3p_ar_detect_app_serial
 91/166 Test  #91: test_deeplabv3p_ar_detect_app_serial .....................   Passed   49.26 sec
        Start  92: test_deeplabv3p_ar_detect_app_mpi
 92/166 Test  #92: test_deeplabv3p_ar_detect_app_mpi ........................   Passed   17.98 sec
```

burlen:
Abdel, I will push a fix for this see #415

elbashandy:
> Abdel, I will push a fix for this see #415

Great. I was able to replicate it on my machine. Setting `mpiexec -n 1` was doubly worse than `-n 2` as well. @burlen I am interested to know the way you debugged and solved this. Did you print the number of threads allocated for each rank?

elbashandy:
app mpi tests are failing only with the serial builds. They're failing when there are >1 ranks per output file. I checked the collective test that we already have `test_cf_writer_collective_mpi`, and is passing normally because it has only 1 rank per output file:

```
start_time_step = 0
end_time_step = -1
n_time_steps = 512
n_time_steps_per_file = 128
n_files = 4
n_ranks = 4
rank	first_time_step	last_time_step
0	0	127
1	128	255
2	256	383
3	384	511
file	ranks
0	0, 
1	1, 
2	2, 
3	3, 
running the test...
STATUS: [0:139928762396928] [/project/TECA/io/teca_cf_reader.cxx:405 TECA-BARD-v1.0.1-34-g2052a46]
STATUS: Found metadata cache ".//556d19f55f71b626b566e4554c4c1694e339c3a4.tmd"
[0:139928762396928] teca_index_executive::initialize index_initializer_key=number_of_tests number_of_tests=1 index_request_key=test_id first=213 last=213 stride=1 block_start=213 block_size=1
[0:139928762396928] teca_index_executive::get_next_request test_id=213 extent=0, 127, 0, 127, 0, 0
STATUS: [0:139928762396928] [/project/TECA/test/test_cf_writer_collective.cpp:313 TECA-BARD-v1.0.1-34-g2052a46]
STATUS: print_info_arrays::execute t=2.61902, s=213, counts=(6200, 10184), z_threshold=0.5
```

Failed tests error (`test_temporal_reduction_app_monthly_average_mpi_thread`):
```
STATUS: [1:140466323400384] [/project/TECA/io/teca_cf_reader.cxx:405 TECA-BARD-v1.0.1-34-g2052a46]
107: STATUS: Found metadata cache "/project/TECA_data/531897a606164b9e34d4ad229189feba841fd5eb.tmd"
107: start_time_step = 0
107: end_time_step = -1
107: n_time_steps = 12
107: n_time_steps_per_file = 7
107: n_files = 2
107: n_ranks = 2ERROR: 
107: rank	first_time_step	last_time_step
107: 0	0	5[1:140466323400384]
107: 1	6	11
107: file	ranks
107: 0	0, 1, 
107: 1	1, 
107:  [/project/TECA/io/teca_netcdf_util.cxx:164 TECA-BARD-v1.0.1-34-g2052a46]
107: ERROR: Collective I/O attempted with a non-MPI NetCDF install. Re-install the NetCDF C library with --enable-parallel-4. See also the output of "nc-config --has-parallel" and/or the value of the "NC-4 Parallel Support" field in the file "<prefix>/lib/libnetcdf.settings" to determine if the install supports parallel I/O.ERROR: [0:140428588521152] [/project/TECA/io/teca_netcdf_util.cxx:164 TECA-BARD-v1.0.1-34-g2052a46]
107: ERROR: Collective I/O attempted with a non-MPI NetCDF install. Re-install the NetCDF C library with --enable-parallel-4. See also the output of "nc-config --has-parallel" and/or the value of the "NC-4 Parallel Support" field in the file "<prefix>/lib/libnetcdf.settings" to determine if the install supports parallel I/O.
107: ERROR: [0:140428588521152] [/project/TECA/io/teca_cf_layout_manager.cxx:96 TECA-BARD-v1.0.1-34-g2052a46]
107: ERROR: failed to create file "prw_monthly_average_%t%.nc"
107: ERROR: [0:140428588521152] [/project/TECA/io/teca_cf_writer.cxx:282 TECA-BARD-v1.0.1-34-g2052a46]
107: ERROR: Failed to create file 0
107: testing on 2 ranksn_threads=2
```

burlen (approved):

PR #378, opened 2020-06-23
Minor bug fixes

### LINKED ISSUE #291: tc_activity test seems to be generating bogus plots
fix

### LINKED ISSUE #374: fix misnomer in teca_metadata_probe --help
`teca_metadata_probe --help` shows the following:
```
  --input_file arg      file path to the simulation to search for tropical cyclones
  --input_regex arg     regex matching simulation files to search for tropical cylones
```

The help should be amended to remove mention of tropical cyclones, since this is a general purpose tool.

### LINKED ISSUE #375: use templates instead of casting as double in teca_metadata_probe
@burlen - I mentioned on Friday that I had a problem running `teca_metadata_probe` on some ERA5 data.

There seems to be a bug in `teca_metadata_probe` that renders it unable to parse ERA5 netCDF files, because they use an integer-valued time coordinate.  The file is readable by TECA, as a python-based pipeline successfully parses the file.  This suggests that there is no problem with the cf_reader itself, but rather there is a problem with the implementation in `teca_metadata_probe`. In attempting to debug this, I noticed that ERA5 uses integer-typed time values.  However, `teca_metadata_probe` is attempting to typecast a double:

[lines 174-181](https://github.com/LBL-EESA/TECA/blob/7aa87914efea17fea342b3b4d1c3ac1ea8d36272/apps/teca_metadata_probe.cpp#L174)
```c++
    teca_metadata coords;
    p_teca_double_array time;
    if (md.get("coordinates", coords)
       || !(time = std::dynamic_pointer_cast<teca_double_array>(coords.get("t"))))
    {
        TECA_ERROR("failed to determine time coordinate")
        return -1;
    }
```

I suspect that the fix would be to wrap this in templating code.  @burlen: could you tackle this please?  I'm only marginally competent with C++ templating, and this would likely take me 10x longer than it would take you.

burlen:
this resolves #374, resolves #375,  and resolves #291

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ env:
     - BUILD_TYPE=Debug
     - TECA_DIR=/travis_teca_dir
     - TECA_PYTHON_VERSION=3
-    - TECA_DATA_REVISION=60
+    - TECA_DATA_REVISION=61
   jobs:
     - DOCKER_IMAGE=ubuntu IMAGE_VERSION=18.04 IMAGE_NAME=ubuntu_18_04
     - DOCKER_IMAGE=fedora IMAGE_VERSION=31 IMAGE_NAME=fedora_31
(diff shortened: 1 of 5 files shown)


taobrienlbl:
Thanks for addressing all of these, Burlen: the changes look good!

Issue #376, opened 2020-06-23
add a non-mpi version of teca library


taobrienlbl:
It would be useful to be able to do `import teca` without requiring MPI calls: particularly when attempting to debug/test python-based pipelines on systems like NERSC, which prohibit using MPI on login nodes.

Ideally, it would be something like:
```python
import teca_nompi

# build TECA pipeline
```

burlen:
It would be easier to have 2 installs. There are some options in the superbuild for instance
```
-DWITHOUT_MPI=ON
```
Tells it not to build any MPI. In that case one would also want to set the install prefix to an alternate location
```
-DCMAKE_INSTALL_PREFIX/some/path/teca_no_mpi
```
then
```
module use /some/path/teca_no_mpi/modulefiles
module load teca
```
to get at the no mpi install. As an alternative one can tell the superbuild to compile MPICH and then one can use MPI on the login node as well.

@taobrienlbl what do you think about the 2 install approach?

burlen:
@taobrienlbl it should be possible to `import teca` from an MPI install of teca when running on a login node. Some of the command line apps do this. For example `teca_convert_table` works fine on the login node.

Things crash when the code also does `from mpi4py import *` since that is what calls `MPI_Init` and Cray makes it abort on the login node.

A work around for the parallel command line apps on the login node might be to use an environment variable `export TECA_NO_MPI=1` (or something like this) which could be tested for before trying to `import mpi4py` or calling `MPI_Init` in the C++ apps.

What do you think?

taobrienlbl:
The two-install approach could work, though I'm somewhat partial to your idea of using an environment variable to control the behavior.  That would allow us to use a single install, which I think would be much simpler from a user point of view.  But if it's difficult to implement, I'm not sure I'd push strongly for this feature.

By the way, I can confirm that simply importing teca does not work on cray:
```bash
$ module load teca
$ python3 -c 'import teca'
[Tue Jun 23 17:40:46 2020] [unknown] Fatal error in PMPI_Init_thread: Other MPI error, error stack:
MPIR_Init_thread(537):
MPID_Init(246).......: channel initialization failed
MPID_Init(647).......:  PMI2 init failed: 1
Aborted
```

[...]

PR #381, opened 2020-06-24
NetCDF I/O avoid MPI calls if MPI is not initialized

burlen:
With this change one can run I/O codes on login nodes (as requested in #376)


burlen:
@taobrienlbl @elbashandy would like  areview, but if you guys are too busy I will merge this

burlen:
resolves #376

burlen:
resolves #399

Issue #392, opened 2020-07-07
use new cropped dataset for all AR tests


burlen:
The new test dataset `ARTMIP_MERRA_2D_2017-05-16-00Z.nc` contains 8 time steps and captures an AR on the North Americal Pacific coast. Replace the old ARTMIP_MERRA time series with this one for all tests that had previously used them.  The This is in TECA_data revision 67.

burlen:
#372 now has updates to use the new dataset in the Bayesian AR detector tests

elbashandy:
Resolved by #372

PR #394, opened 2020-07-08
Replaced ARTMIP datasets that don't have time dimension

elbashandy:
Replaced ARTMIP_MERRA_2D_20170210* with ARTMIP_MERRA_2D_2017-05-16-00Z.nc as the old ones didn't have time dimension. The new one does.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -77,9 +77,9 @@ teca_add_test(test_cf_reader_era5
 
 teca_add_test(test_cf_reader_file_time
     COMMAND test_cf_reader
-    -i "${TECA_DATA_ROOT}/ARTMIP_MERRA_2D_20170210_06\.nc"
+    -i "${TECA_DATA_ROOT}/ARTMIP_MERRA_2D_2017-05-16-00Z\.nc"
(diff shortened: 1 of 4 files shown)


test/CMakeLists.txt
```diff
@@ -77,9 +77,9 @@ teca_add_test(test_cf_reader_era5
 
 teca_add_test(test_cf_reader_file_time
     COMMAND test_cf_reader
-    -i "${TECA_DATA_ROOT}/ARTMIP_MERRA_2D_20170210_06\.nc"
```
burlen, line 80: We need to keep this test as is because the test tests getting time dimension from the file name. In the new dataset the time dimension is encoded in the file itself rather than the file name.

burlen:
See #392

burlen:
@elbashandy  I am updating the Bayesian AR detector tests to use the new data.

elbashandy:
> please update the deeplab test to use the cf reader for the baseline so that the time dimension is tested.

@burlen does the cf_reader read `.bin` format? Or do you mean to insert cf_writer in the pipeline to write a NetCDF file where a cf_reader can read it

burlen:
Abdel will move these changes to #372 when ready

PR #396, opened 2020-07-08
Various fixes and imrpovements

### LINKED ISSUE #395: augment documentation of profiling option
The RTD page on [TECA's profiling capabilities](https://teca.readthedocs.io/en/latest/developer.html#timing-and-profiling) gives the list of environment variables that control the profiler, but it does not specify what values must be provided to activate the profiler.  My first guess, based on the pattern used for activating/deactivating options in TECA's build was that 'ON' should be passed: which is not correct.

Examination of `core/teca_profiler.h` gives more documentation; this should probably be copied verbatim into the RTD page:
```
    // If found in the environment the following variable override the
    // the current settings
    //
    //     PROFILER_ENABLE       : bit mask turns on or off logging,
    //                             0x01 -- event profiling enabled
    //                             0x02 -- memory profiling enabled
    //     PROFILER_LOG_FILE     : path to write timer log to
    //     MEMPROF_LOG_FILE      : path to write memory profiler log to
    //     MEMPROF_INTERVAL      : number of seconds between memory recordings
```

It would also be useful to augment the example shown for profiling the BARD application with the environment variable settings that would activate the profiler in the way shown in Fig 12.

burlen:
a number of minor improvements including:
- resolves #395
- enables streaming mode in the Bayesian AR detector
- other minor enhancements

### DIFF
--- a/alg/teca_bayesian_ar_detect.cxx
+++ b/alg/teca_bayesian_ar_detect.cxx
@@ -984,6 +984,7 @@ const_p_teca_dataset teca_bayesian_ar_detect::execute(
     pr->set_input_connection(pa->get_output_port());
     pr->set_reduce_callback(reduce);
     pr->set_finalize_callback(reduce);
+    pr->set_stream_size(2);
     pr->set_verbose(0);
     pr->set_data_request_queue(this->internals->queue);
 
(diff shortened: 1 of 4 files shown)


burlen:
@taobrienlbl FYI contains the change to use streaming mode in the Bayesian AR detector

burlen:
@taobrienlbl @elbashandy thanks for the review

PR #400, opened 2020-07-09
handle Python 3 ABI flags

### LINKED ISSUE #399: Handle Python3 ABI flags
Python3 introduced ABI flags to differentiate incompatible builds. One such ABI flags is 'm' which indicates use of an optimized allocator.  There are other ABI flags. When compiling against Python3 we need to handle these.  One can discover these from `python-config --abiflags`.

burlen:
resolves #399


burlen:
moving this onto #381, since those fixes are required as well

PR #403, opened 2020-07-14
Fix connected components

taobrienlbl:
Connected components currently counts one too many objects.  Analysis of output meshes indicates that it counts the background as a connected component.  The first commit in this branch adds a test `test/python/test_connected_components_simple.py` demonstrating this on a simple test case.

@burlen - could you please verify that this test is valid and that the failure indeed indicates a bug?

### DIFF
--- a/alg/teca_2d_component_area.cxx
+++ b/alg/teca_2d_component_area.cxx
@@ -87,6 +87,8 @@ void component_area(unsigned long nlon, unsigned long nlat,
     d_cos_phi[0] = calc_t();
     d_cos_phi[nlatm1] = calc_t();
 
+    const component_t neg_one = -1;
+
     // finish off the calc by multiplying the factors
     for (unsigned long j = 1; j < nlatm1; ++j)
     {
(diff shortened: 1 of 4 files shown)


taobrienlbl:
@burlen - I have pushed a proposed fix: the approach labels the background as -1 and starts counting connected components from -1.

Fixes #404

burlen:
This was by design. It makes things easier to treat the background as a component, it's always assigned label 0. for instance the area calculation calculates the area of the background as well. Should we discuss this over zoom?

taobrienlbl:
After discussing this in detail, here is our initial proposal:
Undo commit 01a7a90 in `cc_fix` branch
Improve documentation about the connected components algorithm (in header and RTD)
Rename metadata variable `number_of_components` to `number_of_labels`
Add simple test to ensure that segmentation works as expected if first value of array is not background
Check whether `ar_count` in `teca_bayesian_ar_detect` is reporting AR counts correctly (or whether bg is included)
Consider further how background is/should-be handled with area filters (see [skimage.measure](https://scikit-image.org/docs/stable/auto_examples/segmentation/plot_regionprops.html))

@burlen will add a copy-paste of an e-mail conversation

burlen:
resolved

Issue #404, opened 2020-07-14
connected component bug -- one too many objects counted


taobrienlbl:
See PR #403.

Connected components currently counts one too many objects. Analysis of output meshes indicates that it counts the background as a connected component. The first commit in this branch adds a test test/python/test_connected_components_simple.py demonstrating this on a simple test case.

@burlen - could you please verify that this test is valid and that the failure indeed indicates a bug?

burlen:
The context from our discussions is below:

TAO's comment:
> After discussing this in detail, here is our initial proposal:
> 
> * Undo commit [01a7a90](https://github.com/LBL-EESA/TECA/commit/01a7a90093a983c5897face948e69519f2ae07fd) in `cc_fix` branch
> * Improve documentation about the connected components algorithm (in header and RTD)
> * Rename metadata variable `number_of_components` to `number_of_labels`
> * Add simple test to ensure that segmentation works as expected if first value of array is not background
> * Check whether `ar_count` in `teca_bayesian_ar_detect` is reporting AR counts correctly (or whether bg is included)
> * Consider further how background is/should-be handled with area filters (see [skimage.measure](https://scikit-image.org/docs/stable/auto_examples/segmentation/plot_regionprops.html))
> 

BL's email:

PR #408, opened 2020-07-16
TECA BARD Update

taobrienlbl:
- updates TECA bard parameter table
 - fixed TECA bard count issue documented in #404


taobrienlbl:
We opted to resolve this in #409 instead; closing this PR.

PR #409, opened 2020-07-16
Connected component area background fix

### LINKED ISSUE #404: connected component bug -- one too many objects counted

burlen:
resolves #404
see also #403


burlen:
@taobrienlbl a proposed fix for the issues we have been discussing. please look at it with a very critical eye.

alg/teca_2d_component_area.cxx
```diff
@@ -133,6 +133,8 @@ void teca_2d_component_area::get_properties_description(
         TECA_POPTS_GET(int, prefix, contiguous_component_ids,
             "when the region label ids start at 0 and are consecutive "
             "this flag enables use of an optimization (0)")
+        TECA_POPTS_GET(short, prefix, background_id,
+            "the label id that corresponds to the background (-1)")
```
taobrienlbl, line 137: this default differs from the default set on line 112 - is the idea here that if a user overrides this on the command line, we're assuming that they want to set it to something other than 0?

I'm mainly checking to be sure that -1 isn't a typo.

taobrienlbl: Now I'm wondering if initializing `background_id(0)` is a typo, since it appears this conflicts with the comments provided in the header file:
```
     // By default
     // this is set to -1 to indicate that the value should be obtained from the
     // metadata key `background_id`
```

alg/teca_2d_component_area.h
```diff
@@ -12,24 +12,38 @@ TECA_SHARED_OBJECT_FORWARD_DECL(teca_2d_component_area)
...
+each region. Regions are identified by assigning a unique integer value to each
+mesh point that belongs in the region. The component_variable property names
+the variable containing the region labels.
+
+if the region labels start at 0 and are contiguous then an optimization can be
+used. Set contiguous_component_ids property to enable the optimization. Note that
+TECA's connected component labeler assigns the background (i.e. cells not inside
+the segmentation) the label 0. One can identify the background region and area
+via this label. When processing data generated outside of TECA it might be
+necessary to supply the background label. Use -2 if there is no background.
```
taobrienlbl, line 25: I don't immediately see any logic in the code handling the case of `this->background_id == 2`; all of the tests look for `this->background_id < 0`, in which case -1 wouldn't lead to different behavior form -2.   Is this comment correct, or am I misreading the code?

alg/teca_2d_component_area.cxx
burlen: yes, good catch, it should be -1. I'll make that change now.

alg/teca_2d_component_area.h
burlen: yes, you are right. I'll address this now.

alg/teca_2d_component_area.cxx
```diff
@@ -285,6 +288,20 @@ const_p_teca_dataset teca_2d_component_area::execute(
...
+    {
+        if (in_metadata.get("background_id", bg_id))
+        {
+            TECA_ERROR("Metadata is missing the key \"background_id\". "
+                "One should specify it via the \"background_id\" algorithm "
+                "property")
+            return nullptr;
+        }
+    }
+    out_metadata.set("background_id", bg_id);
```
taobrienlbl, line 303: Should we set `this->background_id = bg_id` after line 303?

Also, it looks like the background id isn't considered in `component_area`; if someone were to pass in a labeled array, where the background is labeled -1 or some other out-of-bounds value, then I think that line 97 would cause a segmentation fault.

I think this could be fixed by changing line 97 to something like
```
// set the index of the current component
int area_index = labels[jj + i];
// if it is the background component, make sure the index is 0
if (area_index == this->background_id) area_index -= this->background_id;
area[labels[jj + i]] += rho_sq_d_theta[i]*d_cos_phi_j;
```

alg/teca_2d_component_area.cxx
```diff
@@ -285,6 +288,20 @@ const_p_teca_dataset teca_2d_component_area::execute(
 
     teca_metadata &out_metadata = out_mesh->get_metadata();
 
+    // get the background_id, and pass it through
+    long bg_id = this->background_id;
+    if (this->background_id < 0)
```
taobrienlbl, line 293: It just occurred to me that this precludes datasets that have the background labeled as a negative value.   This is something that we might actually come across; for example if making use of the '_FillValue' attribute in netCDF, a person might 'remove' the background in a segmented image by setting it to a fill value; this fill value is commonly a negative number like -9999 or -1e-20.  I've actually done exactly this for segmented images before: set the background to -9999 and set the _FillValue attribute to -9999.

One idea--though maybe clunky--could be to have two properties/options:
```
bool override_background_id(false);
long background_id = 0;
```

This would then change the check from `if(this->background_id <0)` to `if ( ! this->override_background_id )`.

alg/teca_component_area_filter.cxx
```diff
@@ -287,6 +287,18 @@ const_p_teca_dataset teca_component_area_filter::execute(
     // get the output metadata to add results to after the filter is applied
     teca_metadata &out_metadata = out_mesh->get_metadata();
 
+    long mask_value = this->mask_value;
+    if (this->mask_value < 0)
```
taobrienlbl, line 291: see the comment in `alg/teca_2d_component_area.h` about this type of check precluding negative values for the mask value.

alg/teca_2d_component_area.cxx
burlen: > Should we set this->background_id = bg_id after line 303?

I don't think so. We don't want to modify the state of the class itself. The only way  `this->background_id` changes is if the user sets it at runtime.

>  the background id isn't considered in component_area

I don't think it needs, but the background id does need to be passed through. & it is.

> if someone were to pass in a labeled array, where the background is labeled -1 or some other out-of-bounds value

In that case the function you reference, the one optimized for contiguous labels starting at 0, can't be used. Instead the implementation that does an indirect lookup must be used. This is controlled by the `contiguous_component_ids` algorithm property.

alg/teca_2d_component_area.cxx
burlen: I have made the change (will share shortly) to use -1 as a code for "get background_id form the metadata", and -2 as a code for "no background label". This preclude the use of -1 and -2 in mask array but would allow any other negative value to be used. How do you feel ab out that?

alg/teca_component_area_filter.cxx
burlen: yup, agree and I've changed to test to `this->mask_value == -1`

alg/teca_2d_component_area.cxx
taobrienlbl: I think that seems okay; I haven't seen any cases where someone might use -1 or -2 as the background label

alg/teca_2d_component_area.cxx
taobrienlbl: got it -- I hadn't caught that an associative array would be passed in; makes sense.  Thanks.

taobrienlbl: sorry, one more thought: should there be a check for consistency between `contiguous_component_ids` and `background_id`?  I someone sets the background to a negative value and sets `contiguous_component_ids`, then I think that could lead to a segfault; might be better to error out if so

burlen: True. and we could scan the label values to see if they are indeed consecutive and start at 0 rather than have a flag that the user would set. I'll make this a new issue though

burlen: see issue #410

PR #414, opened 2020-07-22
CF Reader and Temporal Reduction Modifications for ARTMIP Averaging

taobrienlbl:
* ~~Fixes potential typecast bug in temporal reduction~~
* Adds option to override time attributes/values even when time variable is specified

### DIFF
--- a/io/teca_cf_reader.cxx
+++ b/io/teca_cf_reader.cxx
@@ -935,6 +935,10 @@ teca_metadata teca_cf_reader::get_output_metadata(
                 teca_metadata time_atts;
                 time_atts.set("calendar", this->t_calendar);
                 time_atts.set("units", this->t_units);
+                time_atts.set("cf_dims", n_t_vals);
+                time_atts.set("cf_type_code", int(teca_netcdf_util::netcdf_tt<double>::type_code));
+                time_atts.set("type_code", teca_variant_array_code<double>::get());
+                time_atts.set("centering", int(teca_array_attributes::point_centering));
 
                 atrs.set("time", time_atts);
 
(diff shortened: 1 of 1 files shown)


taobrienlbl:
@burlen - could you please review this PR when you get a moment? 

The type coercion in `teca_temporal_reduction.py` is b/c the calendar utils library expects a double for the time variable, but I encountered a dataset in ARTMIP that uses float for the time variable.

The additonal options in the CF reader are because I want to be able to override the time, calendar, and units attributes (b/c some ARTMIP algorithms have bad data stored in the netCDF variables, so I want to force them to what I know they should be).  Prior to this commit, the `teca_cf_reader.set_t_calendar()`, `teca_cf_reader.set_t_units()`, and `teca_cf_reader.set_t_values()` would only have an effect if the time variable was not set (and therefore not read from the netCDF files).  This revision allows overriding these values while still maintaining the mapping between time indices and filenames inferred by the CF reader when the time variable is set.

burlen:
I'll look at it before cob today

burlen:
before going with this solution I'd like to see if we can get SWIG to automatically convert from numpy scalars to C types. And to see if we could template our code to handle either float or double. I need to verify this but I think we already convert integers in the time axis to floating point.

burlen:
have integrated numpy scalars so that SWIG handles them automatically. more work to do

burlen:
moved numpy & single precision floating point axis code to #442. this PR still has the improvements to the CF reader for overriding the time axis values.

burlen (approved):
these overrides look useful for non CF2 compliant files. I added a commit to integrate with the more recent validation code that checks and alerts for non compliant data and in some cases provides a default.  Things got reordered a bit and I split the calendar and units override such that they are applied separately. Other than that the functionality is as you implemented. If you see any issues with my changes let me know, otherwise please go ahead and merge

Issue #415, opened 2020-07-23
pytorch MPI performance


burlen:
The deeplab ar detector has abysmal performance when run with MPI. This is because of the incorrect massive over subscription of threads by pytorch. When setting to 1 thread the performance is also abysmal and is not viable. We need to expose and use our thread management codes to get the number of cores per MPI rank and forward this information to pytorch. We need to document how to run the pytrroch related codes w/ MPI.

elbashandy:
When you say "When setting to 1 thread the performance is also abysmal and is not viable", do you mean setting pytorch threads to 1 or do you mean using `mpiexec -n 1`?

burlen:
#417 resolved the technical issue, but documentation remains to be completed

burlen:
resolved by #529

PR #417, opened 2020-07-24
Python load balancing

burlen:
* This enables the automatic mapping of the threads to physical cores from Python. This addresses usability issue in `model_segmentation`/`deeplab_ar_detect` when run in MPI parallel the defaults severely degraded performance and it was not obvious why it was occurring.
* runs the deeplab tests in parallel with a fully serial run to compare against.
* includes a number of bug fixes


burlen:
resolves #415

alg/teca_model_segmentation.py
```diff
@@ -69,51 +76,90 @@ def set_pred_name(self, name):
...
+                # we failed to detect the number of physical cores per MPI rank
+                # if this is an MPI job then fall back to 2 threads per rank and
+                # if not let torch use all (what happens when you do nothing)
+                n_threads = -1
+                comm = self.get_communicator()
+                if teca_py.get_teca_has_mpi() and comm.Get_size() > 1:
+                    torch.set_num_threads(2)
+                    n_threads = 2
+                else:
+                    pass
```
taobrienlbl, line 142: Were you intending for the `pass` statement to be a placeholder for other code?  Writing `else: pass` in python is equivalent to not having an else-block at all, which is why I ask.

alg/teca_model_segmentation.py
```diff
@@ -69,51 +76,90 @@ def set_pred_name(self, name):
...
-                "is not a function"
-                )
-
-        if not args:
-            raise ValueError(
-                "ERROR: The provided data transform function "
-                "must at least have 1 argument -- the data array object to "
-                "apply the transformation on."
-                )
+        if threads_per_core > 4:
```
elbashandy, line 97: what is the reason of choosing 4 threads per core here instead of 2 (as represented in the error message)?

alg/teca_model_segmentation.py
burlen: good point, it's not really needed.

alg/teca_model_segmentation.py
burlen: My observation with the test data so far is that 1 thread per core does the best, but the test data is small so I was thinking that 4 would leave room for some experimentation with over subscription. at the same time I wanted to catch the users who didn't realize this is a multiplicative factor. In your performance experiments what did you find?

elbashandy: I found that setting more than 1 thread to torch speeds the process significantly... but the difference in performance of having 2, 4 & 8 threads isn't that big.

I ran on a month worth of data (month of april-2015), 30 files, 8 timesteps per file,  240 timesteps total, On Haswell:

| # of Nodes | # of ranks | # of torch threads | processing time (seconds) |
| ------------- | ------------- | ------------- | ------------- |
| 8  | 240  | 1  | 177 |
| 15  | 240  | 2  | 42 |
| 30  | 240  | 4  | 35 |
| 60  | 240  | 8  | 30 |

alg/teca_model_segmentation.py
burlen: I've now fixed this and a few other issues flagged by pycodestyle.

alg/teca_model_segmentation.py
burlen: you have a physical core available for each thread, this data does not describe performance for over subscription, which is what the setting is related to. if you don';t have that data , not a problem.

elbashandy: Yeah makes sense.

PR #418, opened 2020-07-24
Added a simple TECA-BARD documentation to RTD

elbashandy:


burlen (commented):
This is good start. I think the example on Cori is too complicated for the user guide because it runs the detector in a for loop and uses the file name time template option. We want to include a simple example that includes as few complications as possible. I am working on updating the rtd docs for the Bayesian ar detector to include the new features of computing ivt on the fly and processing high res mip data now. I will cherry pick this commit so you'll get credit for your work, and I will make the needed changes on my branch and merge from there.

burlen:
Thius

doc/rtd/applications.rst
```diff
@@ -666,3 +666,124 @@ regular expression.
         --output_file ${out_dir}/${out_files}                   \
         --arrays PSL TMQ UBOT VBOT U850 V850
 
+Atmospheric Rivers Detection (ARs)
+----------------------------------
+TECA developed detection techniques to quantify the uncertainty of Atmospheric
+River events.
+
+Bayesian - BARD
+---------------
```
burlen, line 675: the wrong section markup (-) was used. Should use ~ here to make this a subsection.

doc/rtd/applications.rst
```diff
@@ -666,3 +666,124 @@ regular expression.
         --output_file ${out_dir}/${out_files}                   \
         --arrays PSL TMQ UBOT VBOT U850 V850
 
+Atmospheric Rivers Detection (ARs)
+----------------------------------
+TECA developed detection techniques to quantify the uncertainty of Atmospheric
+River events.
```
burlen, line 672: This text doesn't introduce the material that follows in this section well. should be revised.

doc/rtd/applications.rst
```diff
@@ -666,3 +666,124 @@ regular expression.
...
+--periodic_in_x FLAG
+    Flags whether the x dimension (typically longitude) is periodic.
+
+Example
+~~~~~~~
+.. code-block:: bash
+
+    teca_bayesian_ar_detect --n_threads 1                            \
+        --input_regex "${data_dir}/ARTMIP_MERRA_2D_2017-05.*\.nc$"   \
+        --output_file test_deeplabv3p_ar_detect_app_output_%t%.nc
```
burlen, line 748: a couple comments about this example:

the output file has deeplab in the name. this might be confusing since this example should illustrate the Bayesian detector (not the deeplab detector).

setting the number of threads explicitly is not recommended. note: teca has internal logic to allocate 1 thread per physical core considering all ranks on a node, this is the default and the recommended approach.

should probably remove this example because running in serial is not practical on most of the datasets of interest.

doc/rtd/applications.rst
```diff
@@ -666,3 +666,124 @@ regular expression.
...
+  do
+      echo "Starting ${year}"
+      srun -n 680 -c 16 -N 40 --cpu_bind=cores teca_bayesian_ar_detect \
+          --input_regex "/global/project/projectdirs/m1517/cascade/external_datasets/ARTMIP/MERRA_2D/${year}/ARTMIP_MERRA_2D_.*\.nc" \
+          --cf_reader::t_axis_variable "" \
+          --cf_reader::filename_time_template  "ARTMIP_MERRA_2D_%Y%m%d_%H.nc" \
+          --steps_per_file 3000 \
+          --cf_writer::date_format "%Y" \
+          --output_file MERRA2.ar_tag.teca_bard_v1.0.3hourly.%t%.nc4 &> bard_${year}_${SLURM_JOB_ID}.log &
+  done
```
burlen, line 784: This is a great example, and shows using the filename_time_template option. We should keep this example as is, and add an additional example of processing HighResMIP data.

burlen (commented):
Commented on a few things that need to be fixed before this goes live

burlen:
Moved these commits onto #510

PR #422, opened 2020-07-30
Catch build errors

burlen:
Set defaults for checking for common dependencies


burlen:
will make  a pass to fix spelling mistakes and typos

PR #433, opened 2020-08-07
Multi cf reader bug fixes

burlen:
a number of critical bug fixes developed in conjunction with the multi-files reader


burlen:
split this into a number of smaller PRs, will close now

PR #435, opened 2020-08-10
Cf reader fixes

burlen:


burlen:
100% tests passed, checks failed because cdash site is down and upload failed

burlen:
> Looks good overall. Could I suggest that we also change the default search order of `metadata_cache_path` to attempt to write to `.` before the CF file directory? It might even be better to have `path` be the first item in the cache search path; since we are increasingly using TECA on community datasets in community directories, it would be better for the default behavior to be to avoid writing in a community directory. (e.g., If I were Bill, I would be mildly annoyed to find that someone had been writing files to the directory structure that I had carefully constructed)
> 
> Suggested change to lines 384-385 of `io/teca_cf_reader.cxx`
> 
> ```c++
> std::string metadata_cache_path[4] =
>             {(getenv("HOME") ? : "."), path, ".", metadata_cache_dir};
> ```
I don't like that idea. because of how netcdf files are layout scanning the time axis is very expensive. I have experienced it taking 10s of minutes on Micahel's when the data set is large. having the cache brings that down to milliseconds when the dataset has already been accessed.  having the file dir first enables the optimization when someone other than you has accessed the dataset. with the high res mip that's even more likely that the cache will optimization will help because multiple people are accessing the dataset. to me it seems this is a strong argument for keeping it the way it is.

What if we hide the files- ie add a dot to the name? this would clear up any clutter.

Can we brian storm some other options?

Either way, I'm hesitant to make the change you are requesting and would like to have it in another PR.

burlen:
@taobrienlbl I added commit (eb26437) just now that prepends . too the cache files. this should reduce the clutter. I agree with you about the cache files being a pain during development, but I also hate to think of the megawatts of power that are wasted when 50k cores are blocked while time axis is scanned. we could potentially redesign the reader to use MPI to scan the time axis and do away with the cache files. It's a non trivial change and would be done in another pr. I created issue #443 as a starting point for this and thinking about other solutions.

taobrienlbl:
I see what you're saying, and making the files hidden seems like an OK compromise for now.  We should follow this discussion up in another thread/issue, as it seems like there is complexity here that is worth hashing out.

PR #436, opened 2020-08-10
array_collection append API

burlen:
adds API for incrementally constructing collections from other collections


burlen:
100% tests passed, checks failed because cdash site is down and upload failed.

burlen:
rebased develop

PR #437, opened 2020-08-10
Fix cf writer type issue

### LINKED ISSUE #370: check type consistency for cf writer
CF Writer needs to check whether teca_variant_arrays are consistent with the typecode given in teca_array_attributes; hard-to-diagnose errors arise if they aren't consistent.

burlen:
resolves #370, among a number of other similar issues relating to types and the CF writer


burlen:
100% tests passed, showing up as failed because cdash site is down and the upload failed.

burlen:
rebased develop

PR #438, opened 2020-08-10
fix cyclic linker dependency between alg and io

burlen:
promote the programmable classes from alg into core. this resolves a linker cycle.


burlen:
100% tests passed but  reported failed because cdash site is down and upload failed

burlen:
rebased develop

PR #439, opened 2020-08-10
fix for bounds to extent when coordinate axis is descending

burlen:
this arises in k hat axis when the units are pressure levels.
the fix is to apply the comparison operator in the binary search bisect
step rather than always using operator<.


burlen:
100% test passed, checks failed because cdash site is down, and submission failed.

burlen:
@taobrienlbl could you give this one a quick look?

PR #442, opened 2020-08-11
Swig wrap numpy scalars

### LINKED ISSUE #426: Handle numpy scalar types in calendar_util and coordinate_util
the temporal average crashes on time values not in double format. the code can be used to test by forcing into float32.

burlen:
* Adds typemaps enabling SWIG to wrap numpy scalars automatically.
* Enables use of single precision floating point time axis.

resolves #426 
moved work from #414


burlen:
100% tests passed, cdash site down, upload failed

burlen:
@taobrienlbl thanks for the corrections

burlen:
rebased from latest develop, will merge if the test suite passes

PR #444, opened 2020-08-12
update cdash drop settings for new nersc config

burlen:
recent cdash outage was caused by migration from one server to another, Jonathan recommends the following changes to get it working again.


burlen:
@elbashandy @taobrienlbl See http://cdash.cdash.development.svc.spin.nersc.org/index.php?project=TECA

PR #449, opened 2020-08-18
Fix cdash build name duplication

### LINKED ISSUE #446: cdash build name duplicated
**Describe the bug**
with the change to 2 runs per os, one for serial netcdf and one for netcdf mpi, the build name was not updated, and as a result the name is duplicated and not easy to tell which run is which.
http://cdash.cdash.development.svc.spin.nersc.org/index.php?project=TECA&date=2020-08-12

**To Reproduce**
run Travis-CI regressions

**Input data**
na

**Platform**
 - HPC system name, if this is a cluster or supercomputer [e.g. Cori Cray XC40]: Travis-CI
 - Operating system name: [e.g. MacOS, Ubuntu, CentOS, etc]: All
 - Compiler and version [e.g. gcc --version]: All
 - CMake version [e.g. cmake --version]: All

elbashandy:
Resolves #446


elbashandy:
I was able to verify it using the travis-ci logs:

ubuntu_serial
ubuntu_mpi


But will rerequest your review once cdash is back.

burlen:
the new name is confusing, because both of these builds do included mpi. how hard would it be to use "netcdf" for serial netcdf and "netcdf_mpi" for a parallel netcdf?

elbashandy:
Good point, will change it to `netcdf` & `netcdf_mpi`

burlen (approved):

PR #450, opened 2020-08-19
[Will remove later] debugging cpu info

elbashandy:


burlen:
@elbashandy 04b58777dda5dbe5261519be598439edf01ff695 looks good, we have to do that anyway until we get cdash situation straightened out. Would you cherry-pick into another pr, and we can merge right away.

burlen:
I think this is debugging code, I will close it for now but feel free to re open if needed

PR #451, opened 2020-08-20
Hide deeplab internals

elbashandy:


alg/teca_deeplabv3p_ar_detect.py
```diff
@@ -401,7 +401,7 @@ def build_model(self, state_dict_deeplab_file=None):
 
         state_dict_deeplab = self.load_state_dict(state_dict_deeplab_file)
 
-        model = DeepLabv3_plus(n_classes=1, _print=False)
+        model = teca_deeplabv3p_ar_detect_internals.DeepLabv3_plus(n_classes=1, _print=False)
```
burlen, line 404: does this still run cleanly through pycodestyle? I guess this line it too long, perhaps there are others

elbashandy: No it doesn't. Cleaning it up now

burlen:
Hi Abdel, the change you wanted to do re the circular dependency would break the PyPi install. I have an alternative  solution in the works.I would still like to get your changes hiding the deeplab internals merged. I've renamed this PR to that affect. Could you please put just these changes into their own commit?

elbashandy:
Just rebased. Thanks @burlen

alg/teca_deeplabv3p_ar_detect.py
```diff
@@ -1,309 +1,314 @@
...
-        self._init_weight()
-
-    def _make_layer(self, block, planes, blocks, stride=1, rate=1):
-        downsample = None
-        if stride != 1 or self.inplanes != planes * block.expansion:
-            downsample = nn.Sequential(
-                nn.Conv2d(self.inplanes, planes * block.expansion,
-                          kernel_size=1, stride=stride, bias=False),
-                nn.BatchNorm2d(planes * block.expansion),
+from teca_py import *
```
burlen, line 7: this seems to be a left over from your experiments with the recursive import. please back this one out too.

burlen (changes requested):

burlen (approved):

Issue #430, opened 2020-08-07
Port the candidate stage of the AEW pipeline


burlen:
Emily's code makes use of a number of general purpose filters including vorticity and Gaussian smoothing that should be factored into algorithms

burlen:
Emily's code is at: https://github.com/mehtut/AEW_Tracker

PR #453, opened 2020-08-20
Wrf reader

### LINKED ISSUE #428: add a WRF reader
To get this off the ground quickly, copy the cf_reader and modify to create `teca_curvilinear_mesh` dataset.
WRF has so called staggered coordinates, which can be used to create node, face, and edge centerings from cell centerings.

Note: Issue #177 has paths to some test data

### LINKED ISSUE #429: Add a curvilinear dataset for use with WRF data
See also #428 and #177

burlen:
* data model improvements for WRF data (arakawa_c_grid, curvilinear_mesh)
* algorithms for arakawa_c_grid (write to VTK, read & write in binary stream format, face to cell interpolation, vertical coordinate transform)
* algorithm for curvilinear_mesh (write to VTK, read and write in binary stream format)
* binary stream error checking and validation
* a regression test exercising wrf reader, face to cell, vertical coordinate transform and write to vtk format.

resolves #428 
resolves #429 

see also #177 , #430

#467 needs to be merged first


alg/teca_dataset_diff.cxx
```diff
@@ -168,86 +170,94 @@ const_p_teca_dataset teca_dataset_diff::execute(
...
-    else
+    else if (dynamic_cast<const teca_arakawa_c_grid*>(ds0.get()))
     {
-        if (this->compare_cartesian_meshes(mesh1, mesh2))
+        if (this->compare_arakawa_c_grids(
+            std::dynamic_pointer_cast<const teca_arakawa_c_grid>(ds0),
+            std::dynamic_pointer_cast<const teca_arakawa_c_grid>(ds1)))
         {
-            TECA_ERROR("Failed to compare cartesian meshes.");
+            TECA_ERROR("Failed to compare arakawa_c_grids.");
```
taobrienlbl, line 251: The error wording here and on lines 241, 231, 221, etc. is ambiguous: does it mean that the comparison failed, as in the comparison could not be done; or does it mean that the comparison yielded a significant difference?  I think it's the latter, in which case an alternate wording might be "Failure because arakawa_c_grids differ"

test/test_array/array.h
```diff
@@ -21,11 +21,14 @@ class array : public teca_dataset
...
     TECA_DATASET_PROPERTY(std::string, name)
 
-
     // return a unique string identifier
     std::string get_class_name() const override
     { return "array"; }
 
+    // return an integer identifier uniquely naming the dataset type
+    int get_type_code() const override
+    { return 1; }
```
taobrienlbl, line 30: This appears to be the typecode for a TECA table, according to `data/teca_dataset_util.h` -- is this correct?  

If not, could you please add some commentary here describing why the number 1 is used and why the typecode isn't determined at runtime (e.g., from the `teca_dataset_tt` struct)? I might not be understanding what this test_array code is used for, and I'm not immediately able to tell from looking at the source.

taobrienlbl (approved):
I've gone through everything and this looks great.  This was a lot of work!  I have added a few minor comments inline.

I have a bit of a concern about the nomenclature that we're adopting: distinguishing 'cf_reader' netCDF files from 'wrf' netCDF files.  But I think that resolving this would involve a longer discussion, and I think the result of that discussion is easily separable from this PR: so no need to hold this up.

Aside from that, I have a couple brief, positive comments about some of the changes. I like the normalization of the environment variable code; that should be useful later on too. And I also like the use of the reader/writer factory method.

test/test_array/array.h
burlen: this is a prototyping code that predates the rest of teca. it shouldn't be referenced from the inside of teca. I'll add a readme file to this directory to make this clear.

alg/teca_dataset_diff.cxx
burlen: Good point. glad to improve the error message.

burlen:
> I have a bit of a concern about the nomenclature that we're adopting: distinguishing 'cf_reader' netCDF files from 'wrf' netCDF files. But I think that resolving this would involve a longer discussion,

So we don't loose track of this ... I have been thinking about how to handle this too, and am so far leaning towards a factory pattern that for instance might examine a file and constructs the appropriate reader based on global attributes or other metadata. But I do have a bunch of unanswered questions about the data model in general. A document like the wrf technical note, but for other data sources, would be extremely helpful in making design decisions. I have a number of data model restructuring changes in the works as well. Look forward to hearing your thoughts and ideas on what we should do.

PR #459, opened 2020-09-03
fix temporal reduction interval iterators

burlen:
Initially 11:59:59 was used for the upper bracket but it should have
been 23:59:59 since times are given in 24 hour format.


burlen:
@taobrienlbl I'm worried that this may change your artmip results.

taobrienlbl (approved):
The changes look good.  It looks like this just shifts the averaging window by 12 hours at most, correct?  Assuming I understood that correctly, I'm not worried about it at all.

burlen:
@taobrienlbl yes, extends the window by 12 hours. for daily average 1/2 the values were left out, less of an impact on the monthly interval. glad that it's not an issue!

PR #464, opened 2020-09-12
Cf reader better calendaring

burlen:
handles the edge case where the time axis in individual files in the data set are given in different units.


taobrienlbl (commented):
The time conversion approach looks correct to me.  Would it make sense to add a small test of both the different-unit feature and the different-calendar error check?

burlen:
I added test data from Paul that has different units in each file , and it is exercised by the temporal reduction tests.

taobrienlbl (approved):
Perfect, thanks Burlen!

PR #467, opened 2020-09-22
table reader CSV

### LINKED ISSUE #324: teca_table_reader read .csv format written by teca table writer
CSV files provide an easy to use path way for ingesting user generated table of data that can be used to provide control data such as lists of files and/or time axis data for the `cf_reader`.

burlen:
Patches so that tables may be created directly from CSV files.

resolves #324


elbashandy (approved):
Had fun reviewing this! gained some insights on stream conversion. Looks great!

taobrienlbl (approved):
These changes look sensible, and I approve the merge.

I have a couple thoughts that might be worth considering for a future revision, both related to the format of CSV file that the table reader can parse.  I noticed in `data/teca_table.cxx` that some aspects of the CSV file are assumed/hardcoded: comment lines begin with '#', and string data are indicated by double quotes not single quotes.  This seems fine for the current use, where the CSV files are (presumably) ones generated by TECA and are in this format. If it is intended that the CSV table reader can handle externally-generated CSV files, then it might be good to revise the code so that the comment string and string-delimiter can be provided as options.  Some additional options might also be necessary, since table files sometimes have 'implied' headers: no header string is used, but its assumed that your code will skip the first N rows.  The documentation for [`pandas.read_csv()`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_csv.html) might provide a good reference for the types of options that might need to be implemented in order to be able to parse the variety of CSV formats that occur in the wild.

Alternatively/additionally, it might be useful to provide a short example (e.g., in a block-comment in `data/teca_table.h`) of a CSV file that TECA is known to be able to parse.

burlen:
added CSV format documentation to teca_table_reader header.

PR #476, opened 2020-09-28
App tests linking/mpi rank/diff order fixes

### LINKED ISSUE #468: order of inputs to teca_table_diff
**Describe the bug**
the order of the command line arguments to teca_table_diff are reversed compared to teca_dataset_diff. This is confusing and break common convention of specifying the reference file first.

**Describe the solution**
re-order the command line arguments such that the reference data is supplied first and the test data second.

### LINKED ISSUE #471: app tests hard coded to use 2 mpi ranks
**Describe the bug**
The app tests are hard coded to use 2 mpi ranks.

**Describe the solution**
For most of the test the mpi concurrency should be based on the cmake variable(s) TEST_CORES and/or HALF_TEST_CORES so that when there are more than 2 cores the tests run faster and MPI parallelism is stressed. Exceptions include when the test data has 2 or fewer time steps, when threading concurrency is automatically determined, and where there are other special considerations.

### LINKED ISSUE #472: App tests linking spec
**Describe the bug**
it is not necessary to specify LIBS when defining app tests in CMake.

**Describe the solution**
The app tests are based on shell scripts that exercise the pre-compiled or python applications, there is no compile and link step. the LIBS spec should be removed.

### LINKED ISSUE #473: order of inputs to teca_cartesian_mesh_diff
**Describe the bug**
the order of the command line arguments to teca_table_diff are reversed compared to teca_dataset_diff. This is confusing and break common convention of specifying the reference file first.

**Describe the solution**
re-order the command line arguments such that the reference data is supplied first and the test data second.

elbashandy:
Resolves #471 
Resolves #472 
Resolves #468 
Resolves #473

### DIFF
--- a/apps/teca_cartesian_mesh_diff.cpp
+++ b/apps/teca_cartesian_mesh_diff.cpp
@@ -24,7 +24,7 @@ int main(int argc, char **argv)
         if (rank == 0)
         {
             std::cerr << std::endl << "Usage:" << std::endl
-                << "teca_cartesian_mesh_diff [test file] [reference file]"
+                << "teca_cartesian_mesh_diff [reference file] [test file]"
                 << std::endl << std::endl;
         }
         return -1;
--- a/apps/teca_table_diff.cpp
+++ b/apps/teca_table_diff.cpp
@@ -28,7 +28,7 @@ int main(int argc, char **argv)
         {
             cerr << endl
                 << "Usage:" << endl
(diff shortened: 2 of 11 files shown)


burlen:
Hi Abdel, Would you please rebase? I'll review then

elbashandy:
Just rebased!

burlen (approved):

PR #479, opened 2020-10-01
WIP - Cartesian mesh refactor

burlen:
work improving dataset API now that we have multiple types of Cartesian meshes


burlen:
closing because this is stale. changes look OK, but are not in use anywhere.

PR #480, opened 2020-10-02
Integrated vapor transport

### LINKED ISSUE #390: Vertical integral algorithm
We need an algorithm for calculating vertical integrals: both on native model levels and on pressure levels.  #354 will close this.

This may depend on #353, since the pressure level calculation may need surface variables (e.g., surface specific humidity) for an accurate calculation; these are usually stored in separate files from height-dependent versions.

burlen:
* adds teca_vertical_integral a base class containing common code for transforming meshes from 3D to 2D by a reduction along the vertical axis.
* adds teca_integrated_vapor_transport an instance of a teca_vertical_reduction that calculates IVT from specific humidity and wind fields. includes a test validating the calculation against an analytically derived result. 

resolves #390.

### DIFF
--- a/alg/CMakeLists.txt
+++ b/alg/CMakeLists.txt
@@ -43,6 +43,7 @@ set(teca_alg_cxx_srcs
     teca_tc_trajectory.cxx
     teca_temporal_average.cxx
     teca_variant_array_operand.cxx
+    teca_vertical_reduction.cxx
     teca_vorticity.cxx
     teca_dataset_diff.cxx
     )
--- a/alg/CMakeLists.txt
+++ b/alg/CMakeLists.txt
@@ -25,6 +25,7 @@ set(teca_alg_cxx_srcs
     teca_descriptive_statistics.cxx
     teca_evaluate_expression.cxx
     teca_geography.cxx
+    teca_integrated_vapor_transport.cxx
     teca_l2_norm.cxx
     teca_latitude_damper.cxx
     teca_laplacian.cxx
--- a/alg/teca_integrated_vapor_transport.cxx
+++ b/alg/teca_integrated_vapor_transport.cxx
@@ -1,0 +1,326 @@
+#include "teca_integrated_vapor_transport.h"
+
+#include "teca_cartesian_mesh.h"
+#include "teca_array_collection.h"
+#include "teca_variant_array.h"
+#include "teca_metadata.h"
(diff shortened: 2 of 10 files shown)


burlen:
@taobrienlbl I have IVT for Cartesian meshes implemented. Should we (pending review) merge this now so that folks who have data in an amenable form can proceed while we tackle the coordinate transforms you mentioned in #390? Would also be a small victory to report this progress next week.

taobrienlbl:
@burlen, yes, I think this would be good to merge in pending review.

On an initial look through the code, I have one immediate comment: the `teca_integrated_vapor_transport` code should be refactored and renamed, to `teca_vertical_integral`; we'll be using this for more than just calculating IVT (though that is our first target).  The integration code itself looks great; my comment is that the IVT part (below) should be factored out of the integration routine, and the integration routine should allow an arbitrary 4D array as input:

```c++
    // compute the integrand
    num_t *f = (num_t*)malloc(nxyz*sizeof(num_t));
    for (unsigned long i = 0; i < nxyz; ++i)
        f[i] = wind[i]*q[i];
```

I think the above could be processed more generically using TECA's math parsing code (`teca_parser.cxx`, I think), right?

burlen:
> the `teca_integrated_vapor_transport` code should be refactored and renamed, to `teca_vertical_integral`; we'll be using this for more than just calculating IVT

I'm working on such an algorithm(not yet shared) where one specifies for _each_ integrand :

1. a string containing a mathematical formula defining the integrand
2. a similar formula to applied post integration, for the application of factored out constants etc
3. a string containing name of the result
4. array attributes for the result.

While we can hold off until this is finished, I'll be  working on another project for the next few weeks. With what we have in hand now we could deploy a simple app fairly quickly, if the IVT calcs become really urgent.

I think we might keep a specific IVT implementation around, at least in the short term. Higher performance of the specialized implementation is one reason. It gives the ability to compare performance of the parser based solution to something specific. I'm curious to see how much faster the specialized implementation is. I'm thinking about performance comparisons to other tools as well. The specialized implementations are also slightly easier to use than the general purpose solution.

taobrienlbl:
@burlen, got it-- that makes sense to move forward with this version then. Thanks.

taobrienlbl (approved):

burlen:
I think something is messed up with this pr on github side, not syncing up. I'll try to create a new one

PR #484, opened 2020-10-06
Cartesian mesh reader regex support

### LINKED ISSUE #288: add regex support to teca_cartesian_mesh_reader
this would enable the reader to process time series.

motivation for this feature is primarily for regression testing. we currently don't have a way to to test across the time domain

We should be able to leverage same code that's used in the cf reader

### LINKED ISSUE #316: support time series baselines
regression tests currently can only run on one time step because the teca_cartesian_mesh reader doesn't handle time series. enabling this feature requires adding regex parsing to the cartesian mesh reader and may require massaging the request generation in the dataset diff.

elbashandy:
Resolves #288 
Resolves #316


burlen (commented):
a couple of changes to make here.

PR #495, opened 2020-10-23
Allow run-time setting of row dimension name for teca_table_writer netCDF output

taobrienlbl:
Allow run-time setting of row dimension name for teca_table_writer netCDF output


burlen (approved):
these look fine to me

taobrienlbl:
thanks @burlen !

PR #496, opened 2020-10-23
Bayesian ar app update

burlen:
improvements to the teca_bayesian_ar_detect app

* use the multi_cf_reader to enable processing of HighResMIP data
* adds command line options to compute and write IVT
* adds command line options to compute and write IVT magnitude
* adds a new test for the new capabilities


taobrienlbl (approved):
This looks clean!  I really appreciate you doing this--I was mulling over how to tackle this, and I really appreciate seeing this implementation.  Also, that's super cool that you wired in the IVT integration.  That'll be immediately beneficial!

My one suggestion for a change here would be to add the example `.mcf` file that is in sourceforge to the docs (maybe to the doc for the bayesian AR detector app?).

burlen:
> My one suggestion for a change here would be to add the example `.mcf` file that is in sourceforge to the docs (maybe to the doc for the bayesian AR detector app?).

yes, we definitely need that. I've started a new chapter in the rtd docs where there is a section for each teca algorithm, this would go in there. I think we might also want a chapter titled "Getting data into and out of TECA"  which would be a tutorial with examples of readers and data model.

burlen:
The test is failing on travis-ci very likely because this data is too large to process on 1 core. I'll work on that before merge

PR #501, opened 2020-11-06
Update more apps for highresmip

burlen:
* updates tc_detect app for HighResMIP data
* updates tc_wind_radii app for HighResMIP data
* updates cf_restripe app for HighResMIP data
* updates cartesian_mesh_diff app for HighResMIP data
* adds verbose mode to dataset_diff algorithm so that it's obvious what is being tested
* adds report about available variables/arrays to the metadata probe app


burlen:
@taobrienlbl fyi

PR #506, opened 2020-11-14
Cf reader performance

burlen:
MPI parallelize reading the time axis in the cf reader.

The major contributions of this PR are an MPI parallel map reduce pipeline that is used inside the `teca_cf_reader`. New classes include `teca_cf_time_axis_data`, `teca_cf_time_axis_reader`, and `teca_cf_time_axis_data_reduce`. Pipeline internals were updated to support running on a subset of ranks by partitioning the MPI communicator.

By default at most 1024 ranks will be used to read the time axis. This works well on Cori but may not be optimal on other systems. The max number of reader ranks can be adjusted via algorithm properties or on the command line.

the number 1024 was arrived at by running the meta_data_probe on a 7300 file CAM5 dataset with different numbers of ranks. in the runs all ranks were used to read the time axis. With one day per file the large number of files in this dataset puts it at the extreme end of the spectrum. Strong scaling starts to tale off around 1024 ranks. See the below plot.

mdp_performance_7300_files


burlen:
mdp_performance_7300_files

PR #517, opened 2020-12-02
Seasonal reduction

burlen:
add seasonal interval for DJF, MAM, JJA, SON


burlen:
cdash.nersc.gov is down due to power outage at nersc, this is resulting in TravisCI builds timing out. will merge this now

PR #519, opened 2020-12-10
Bug fixes develop

burlen:
* acquire gil
* rename shared temporal reduction baselines


burlen:
the deeplab test fail because of baseline name change in another branch. will merge this now,

PR #565, opened 2021-02-28
Masking in teca bard

taobrienlbl:
Adds the ability to mask input variables when running TECA BARD: fields are multiplied by `ar_probability`.  The use case here is calculating quantities like AR precipitation.


alg/teca_apply_binary_mask.cxx
```diff
@@ -30,6 +54,94 @@ teca_apply_binary_mask::teca_apply_binary_mask() : mask_variable("")
...
+        }
+        else
+        {
+            // copy the attributes from the input. this will capture the
+            // data type, size, units, etc.
+            teca_array_attributes output_atts(input_atts);
+
+            // update description.
+            output_atts.description = 
+                std::string("masked/weighted by `" + this->mask_variable + "`");
```
burlen, line 135: we should also update the long_name

alg/teca_apply_binary_mask.cxx
```diff
@@ -30,6 +54,94 @@ teca_apply_binary_mask::teca_apply_binary_mask() : mask_variable("")
...
+{
+#ifdef TECA_DEBUG
+    cerr << teca_parallel_id()
+        << "teca_apply_binary_mask::get_output_metadata" << endl;
+#endif
+    (void)port;
+
+    if (this->input_variables.empty())
+    {
+        TECA_WARNING("The list of input variables was not set")
```
burlen, line 103: might be good to warn only on rank 0 for large parallel runs

burlen (approved):
I saw a few things, most of them minor. the main points were that by processing teca_mesh(instead of teca_cartesian_mesh) we could handle all mesh types. Also the code style (eg braces) was not matching ours completely. Rather than complain at you I made a pass over the code to address these issues and added an ar weighted variable to the bayesian_ar_detect_app_test. I'll push these now

burlen:
If you're curious to see what I changed you could in the "files changed tab" view "changes from" menu in the top left to select the individual commits I made.

taobrienlbl:
Thank you @burlen for manually fixing these issues!  Glad to see it otherwise seemed reasonable.  I do need to ask you at some point about any command(s) you use to check style.

burlen:
oops, my changes broke your new test, will push fix now

PR #569, opened 2021-03-05
Adds a dataset caching algorithm

burlen:
The teca_indexed_dataset_cache algorithm caches datasets using the index_request key value from the request. The cache size has to be set at run time. When the cache size is exceeded the least recently used dataset is removed from the cache. This functionality is key to reducing I/O overheads when the same data is accessed repeatedly. A couple of use cases currently exist.
1. serving surface elevation field when masking mesh points where the vertical pressure coordinate falls below the surface
2. serving land sea mask
3. computing a simple moving average over the time dimension
4. feature tracking where multiple previous time steps are needed for each current time step, eg to compute feature overlap


taobrienlbl:
Another great use-case is training algorithms like the Bayesian AR detector, where parameter settings are repeatedly tested on the same input field.  I/O ended up being a significant cost for training TECA BARD.  This is fantastic!

burlen:
I'd be very excited if TECA could be used to train BARD more efficiently than otherwise possible, assuming that re-training is ever needed/useful.

PR #580, opened 2021-03-22
Teca fix documentation 1

elbashandy:
- Added/organized the brief descriptions to all classes. Most of the classes already had brief descriptions but weren't being picked up by Doxygen.
- Excluded the unnecessary files like `core/teca_type_select.h`


burlen (commented):
Having something show up in a table in doxygen page is not the goal.

 **_The goal is for the brief documentation to describe the purpose of the class concisely in a couple/few sentences such that someone unfamiliar with TECA, but familiar with climate science, could find the teca_algorithm classes needed to build a pipeline to accomplish what ever analysis task they are trying to accomplish._**

What I see here is largely shuffling around the existing, flawed, inadequate, documentation.  I don't see the point in simply shuffling things around with out making the needed revisions and improvements as I will have to make pass over it anyway to make the needed revisions. 

I have a few requests as you continue to work on this:

Please do not add documentation that does not describe the class. Eg, what you did with teca_bad_cast, teca_output_port, teca_threadsafe_queue etc. I also made a similar request in #505.

When revising, please fix any issues you come across as you go. please use full sentences with capitalization and punctuation. Fix spelling and grammatical mistakes,  etc.

Please adopt the conventions for doxygen comments that I shared with you in review of #505. Not following this convention will make more work us, and what's the point of doing this 2 times? I'll repeat the example from #505 here with some refinements based on recent experiences.

For classes that only have a one line brief description, it is OK to use:
```
/// This is a brief description.
class foo
{
};
```
If the class can be briefly described in a single sentence and has a detailed description, it is OK to use:
```
/// This is a brief description.
/**
 * More details go here.
 */
class foo
{
};
```
For classes that have only a brief description that takes more than one line, you would have to use:
```
/** @brief
 * This is the brief description (pretend this extends 80 chars)
 * spanning multiple lines.
 */
class foo
{
};
```
For classes that have brief description spanning multiple lines and a detailed description, you would have to use:
```
/** @brief
 * This is the brief description.
 *
 * @details
 * These are the details.
 * /
class foo
{
};
```
This latter is what I've settled on in my own revisions. For some real examples already in the repo see also :

https://github.com/LBL-EESA/TECA/blob/develop/alg/teca_valid_value_mask.h
https://github.com/LBL-EESA/TECA/blob/develop/alg/teca_cartesian_mesh_regrid.h


When revising, comments spanning multiple lines should wrap the lines < 80 chars, and ideally would not span multiple lines where they would fit on one.  In vim this is simple, highlight the text and hit "gq"

All of the above are things I'm striving for in my own revisions. I want us to be on the same page and working together toward the same goal so that your efforts aren't wasted and have the most impact.

taobrienlbl:
@elbashandy - could you please trigger RTD to build this branch and post the link here?

elbashandy:
@burlen I chose to focus on the technical aspect of the documentation first before diving into the long-term vision of improving the overall inadequate documentation of TECA. From our conversations, we decided that I wasn't going to tackle the details of the documentation (yet) as I am already unfamiliar with parts of TECA (e.g teca_bad_cast). My focus was to have Doxygen operate correctly before diving into the semantics of the TECA components.

To avoid misunderstanding, I created few issues that @taobrienlbl and I had discussed beforehand about how we are going to tackle the documentation overall; and I added these issues to the aforementioned project that we talked about last week (Project name: Documentation). Your feedback regarding these issues is appreciated.

Thanks for pointing out your previous request regarding the commenting styling convention that I am going to apply to all classes. I will change the documentation accordingly.

elbashandy:
@taobrienlbl https://teca.readthedocs.io/en/teca_fix_documentation_1/api/generated_rtd_io.html

burlen:
I don't expect nor want you to add documentation to classes you don't understand, that would create more problems than it would solve. my point about teca_bad_cast and the others is don't add a useless comment.

Since this stuff is going live on our site when it's merged, I think it is an entirely reasonable request that, as you are in there inserting doxygen style comment markers and you have eyes on the code, to at the same time 1. Fix spelling mistakes (I use vim ":set spell spelllang=en_us"), 2. Capitalize, punctuate, make complete sentences out of fragments, and 3. rewerap at 80 chars (vim select and "gq"). I'm only talking about the lines you're already modifying. I would assume that these type of edits would be quick and easy to do. This would overall be very helpful and move us closer to the goal.

burlen (approved):
I see many fixes and improvements. As per previous reviews on this pr and others, please don't add comments that aren't documentation (eg "todo document x"). Also some sections were reformatted that didn't need to be. At any rate, I think it's best for you to merge this now.

elbashandy:
@burlen The reason I added a "TODO: ..." is to show that we are aware that the documentation is missing and we are active about it... compared to the blank entries which look buggy, inconsistent, and does not indicate any activity that we are aware of them.

I have a question: Is there a hard limit on the number of characters that we allow in a line. Previously, you have requested comments to be < 80 characters. The reason I changed some of these comments is that they are >= 80 characters. I added a ruler in my IDE to show me when I pass 79 characters:

> When revising, comments spanning multiple lines should wrap the lines < 80 chars, and ideally would not span multiple lines where they would fit on one. In vim this is simple, highlight the text and hit "gq"

burlen:
you're pr is approved, and I'd hoped you'd merge it and move on rather than pick a fight.

I am the technical lead on this project and you are a junior engineer. you asked me for a code review and I gave it. when I ask you for something there is a reason, you may not like it, it may not be what you'd do, but you should respect and honor those requests without arguing.

PR #593, opened 2021-04-06
Adding extra documentation to parser_xml.py

elbashandy:


taobrienlbl (changes requested):
As discussed in person today: the added docstrings are great.  This PR should also include comments on the main code (e.g., avoid having long blocks of code w/o comments).

PR #601, opened 2021-04-28
Cf writer interval bug

burlen:
* fixes off by one in monthly layout
* adds more verbose diagnostic info to the writer and tests
* fixes an issue where some internal control attributes were being written. when read these were overriding the runtime settings leading to inconsistency when 3D dataset was sliced. skipped writing them and restructured so that run time setting soverride any found on disk.


burlen:
cvlosing this to see if I can work around travis-ci setup issue that should now be resolved

PR #625, opened 2021-06-03
WIP -- generate valid value masks from surface pressure

### LINKED ISSUE #526: pressure from sea level pressure
For the IVT integral in highresmip, we should mask values that are below the ground surface.  Most models have sea level pressure (variable name `psl`); with a surface elevation field, surface pressure can be estimated.  The condition for valid data is then $p >= ps$, where `ps` is the surface pressure.

A particular model has both ps and psl, which should facilitate testing:

PSL:
/global/cfs/cdirs/m3522/cmip6/CMIP6_hrmcol/HighResMIP/CMIP6/HighResMIP/ECMWF/ECMWF-IFS-HR/highresSST-present/r1i1p1f1/6hrPlevPt/psl/gr/v20170915/

PS:
/global/cfs/cdirs/m3522/cmip6/CMIP6_hrmcol/HighResMIP/CMIP6/HighResMIP/ECMWF/ECMWF-IFS-HR/highresSST-present/r1i1p1f1/Prim6hrPt/ps/gr/v20170915

burlen:
resolves #526

### DIFF
--- a/alg/teca_surface_pressure_estimate.cxx
+++ b/alg/teca_surface_pressure_estimate.cxx
@@ -1,0 +1,262 @@
+#include "teca_surface_pressure_estimate.h"
+
+#include "teca_mesh.h"
+#include "teca_array_collection.h"
+#include "teca_variant_array.h"
+#include "teca_metadata.h"
+#include "teca_array_attributes.h"
+
+#include <algorithm>
+#include <iostream>
+#include <string>
+#include <set>
+#include <cmath>
+
+#if defined(TECA_HAS_BOOST)
+#include <boost/program_options.hpp>
+#endif
+
+//#define TECA_DEBUG
+
+namespace internal
+{
+template <typename num_t, typename coord_t>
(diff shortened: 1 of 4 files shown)


alg/teca_surface_pressure_estimate.cxx
```diff
@@ -0,0 +1,262 @@
...
+
+//#define TECA_DEBUG
+
+namespace internal
+{
+template <typename num_t, typename coord_t>
+void compute_surface_pressure(size_t n_elem, const num_t *tas,
+    const num_t *psl, coord_t *z, num_t *ps)
+{
+    num_t cp = 1.0; // kJ/kg.K
```
taobrienlbl, line 27: cp and cv should be in units of J/Kg/K.  This would matter for `gam` since the units cancel, but it does matter for the `GAM * zi` parts.

I suggest `cp=1004.` and `cv = 718.` as the values used

PR #646, opened 2021-09-02
Added TECA-related publications

taobrienlbl:
@burlen - any objections to keeping a list of TECA-related pubs in `doc/TECA_publications.md` ? I took a subset from the list you sent me: papers with URLs available.


burlen:
@taobrienlbl I have no objections to this

PR #650, opened 2021-09-21
revise error handling

### LINKED ISSUE #648: python exceptions aren't raised after #645
Both @indahector and I are developing python-based TECA pipelines and have found that information about Python's exceptions is no longer printed as of #645 .  We have both been using the stable build at NERSC while debugging.

Is there some sort of runtime configuration that print Python exception information?

### LINKED ISSUE #649: TECA_ERROR abort refinement
As a follow up to #645. We would add a `TECA_FATAL_ERROR` macro that aborts and let `TECA_ERROR` macro return. `teca_algorithm` subclasses would be updated to use `TECA_FATAL_ERROR` and various utility libraries would use `TECA_ERROR`. That way we'd still get info about what was running when the error occurred.

burlen:
adds TECA_FATAL_ERROR that invokes the errorhandler (the default will abort). This is only used in pipeline execution phases, elsewhere TECA_ERROR is used. This will allow more contextual information to be reported before the abort.

updates the Python error reporting macros to remove ANSI color coding and flush error and output streams.

a follow up revision to #645 
resolves #648
resolves #649

### DIFF
--- a/alg/teca_2d_component_area.cxx
+++ b/alg/teca_2d_component_area.cxx
@@ -198,7 +198,7 @@ std::vector<teca_metadata> teca_2d_component_area::get_upstream_request(
     std::string component_var;
     if (this->get_component_variable(component_var))
     {
-        TECA_ERROR("component_variable was not specified")
+        TECA_FATAL_ERROR("component_variable was not specified")
         return up_reqs;
     }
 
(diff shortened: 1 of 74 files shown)


burlen:
@taobrienlbl if it is easy to do, would you be able to see if this resolves #648?

taobrienlbl (approved):

PR #669, opened 2021-11-08
WIP -- time_axis_convolution

burlen:
Adds a new algorithm to compute kernel convolution along the time axis.
Support for two predefined kernels (Gaussian, and constant) as well as for explicitly
provided kernels is implemented. The predefined Gaussian kernel is initialized from user
provided center, max height, and width.  Includes a regression test and
Python bindings.


burlen:
@taobrienlbl  Here is a draft of the time axis convolution. There is a regression test that illustrates its use, but this will need some more validation.

taobrienlbl:
@burlen  - that's awesome!! Thanks for pulling this together so quickly!

It looks like `test_time_axis_convolution.cpp` might not have been committed -- at least I don't see it in the file tree.  Maybe I'm overlooking it?

burlen:
@taobrienlbl I've finished validating and have cleaned this up. There are now 3 regression tests. I will merge what's here into the develop branch. You should be all set now to add the kernels you need.

By the way, the regression tests can be used as a command line app to apply the convolution and write the data in parallel using threads. Setting TECA_DO_TEST=0 in the environment will run it this way. This was useful during development for dumping and then visualizing the results. 

```
TECA_DO_TEST=0 ./bin/test/test_time_axis_convolution  \
    "/home/bloring/work/teca/TECA_data/prw_hus_day_MRI-CGCM3_historical_r1i1p1_19500101-19501231\.nc$"  \ 
    "./test_time_axis_convolution_backward_constant" "constant" "3" "backward" "0" "-1" "-1" "prw"
```

Let me know what else I can do to help.

PR #676, opened 2021-12-07
Added test_lapse_rate.py

amandasd:
It is just a test. I am not asking for merging into develop branch.


amandasd:
It is just a test. I am not asking for merging into develop branch.

burlen:
@amandasd there was an issue with the build. the place to look for more info is cdash.nersc.gov. here is a link. https://cdash.nersc.gov/index.php?project=TECA

You'll want to click on the red entry in the "tests failed" column, then click on the failed test in the "status" column to see the test output.

test/python/CMakeLists.txt
```diff
@@ -481,17 +481,16 @@ teca_add_test(py_test_config
...
 teca_add_test(py_test_thread_parameters_serial
     COMMAND ${PYTHON_EXECUTABLE}
-    "${CMAKE_CURRENT_SOURCE_DIR}/test_thread_parameters.py"
-    FEATURES ${MPI4Py_FOUND})
+    "${CMAKE_CURRENT_SOURCE_DIR}/test_thread_parameters.py")
 
 teca_add_test(py_test_thread_parameters_mpi
     COMMAND ${MPIEXEC} -n ${TEST_CORES} ${PYTHON_EXECUTABLE} -m mpi4py
-    "${CMAKE_CURRENT_SOURCE_DIR}/test_thread_parameters_mpi.py"
+    "${CMAKE_CURRENT_SOURCE_DIR}/test_thread_parameters.py"
```
burlen, line 488: why do we need that?

alg/teca_lapse_rate.py
```diff
@@ -0,0 +1,168 @@
...
+    Uses 1st order finite difference to estimate the lapse rate at all available levels and then averages. Levels
+    below the surface are masked, and levels above zmax are masked.
+    
+    It is assumed that the vertical dimension is the first dimension (e.g., [level, lat, lon]) for both t and z.
+    
+    """
+    
+    # check shapes
+    assert t.shape == z.shape, "t and z must have the same shape; shape(t) = {t.shape} but shape(z) = {z.shape}"
+    assert z[0,...].shape == zs.shape, "All but the first dimension of z must match the dimensions of zs"
```
burlen, line 32: It is OK to leave the assert if this is only for debugging, but if this is a condition that is likely to occur in practice then we probably should always test and raise an exception.

alg/teca_lapse_rate.py
```diff
@@ -0,0 +1,168 @@
...
+    
+    # get a boolean mask for under-topo values and values that are above the max height
+    height_mask = np.logical_or(z < zs[np.newaxis,...], z >= zmax)
+    
+    # mask temperature and height
+    t_mask = np.ma.masked_where(height_mask, t)
+    z_mask = np.ma.masked_where(height_mask, z)
```
burlen, line 39: cupy doesn't implement Numpy's masked arrays so this will need to change. 

Consider using "boolean indexing" to directly select elements. A silver lining is that boolean indexing is more efficient and generally faster than `numpy.were` because it selects elements directly rather first building a list of indices.

burlen (commented):
@amandasd Nice work. I made a few comments in line that you should take a look at. additionally, would you please run `pycodestyle` on the new python classes and fix issues it reports?  Over all things are looking good!

alg/teca_lapse_rate.py
```diff
@@ -0,0 +1,168 @@
+import numpy as np
+
+def calculate_lapse_rate(t,z,zs,zmax = 9000):
```
burlen, line 3: please move this into the `teca_lapse_rate` class. as is it will be introduced at module scope, but it's really an implementation detail that need not be exposed and will keep the module cleaner and easier to use to encapsulate this in the class that uses it.

alg/teca_lapse_rate.py
```diff
@@ -0,0 +1,201 @@
...
+        z_mask = np.where((z >= zs[np.newaxis, ...]) & (z < zmax), z, np.nan)
+
+        # calculate the lapse rate between each level
+        dt = np.diff(t_mask, axis=0)
+        dz = np.diff(z_mask, axis=0)
+
+        # calculate the average lapse rate
+        dtdz = np.nanmean((dt/dz), axis=0)
+
+        return dtdz
```
burlen, line 102: In this and some of the other functions you're using numpy all the time (via a global variable `np`). You'll need to select either numpy or cupy in each function and assign to a local variable. Note, what you've done in `execute` is right, you'll need to do this in the other functions as well.

burlen (commented):
Looking good, you'll need to make sure you're using cupy in all the functions.

alg/teca_lapse_rate.py
```diff
@@ -0,0 +1,201 @@
...
+        masked, and levels above zmax are masked.
+
+        It is assumed that the vertical dimension is the first dimension
+        e.g., [level, lat, lon]) for both t and z.
+
+        """
+
+        # check shapes
+        try:
+            assert t.shape == z.shape, "t and z must have the same shape; \
```
burlen, line 79: don't use assert to check for errors that are expected to occur post development because it can be compiled out. instead replace with an appropriate conditional. also don't catch the exception if this is something that cannot be ignored

alg/teca_lapse_rate.py
burlen: this one didn't get addressed yet, see newer comment on that piece of code

alg/teca_python_vertical_reduction.py
```diff
@@ -62,6 +64,18 @@ def request(self, port, input_md, request):
...
+        dev = -1
+        np = numpy
+        if get_teca_has_cuda() and get_teca_has_cupy():
+            dev = request['device_id']
+            if dev >= 0:
+                cupy.cuda.Device(dev).use()
+                np = cupy
+        # report
+        dev_str = 'CPU' if dev < 0 else 'GPU %d' % (dev)
+        sys.stderr.write('teca_python_vertical_reduction::request %s\n' % (dev_str))
```
burlen, line 77: to avoid a massive amount of text when running in parallel use 
```
if self.get_verbose():
     # code to print status message
```

alg/teca_python_vertical_reduction.py
```diff
@@ -62,6 +64,18 @@ def request(self, port, input_md, request):
...
         md = teca_metadata(input_md[0])
 
+        # get the device to run on
+        dev = -1
+        np = numpy
+        if get_teca_has_cuda() and get_teca_has_cupy():
+            dev = request['device_id']
+            if dev >= 0:
+                cupy.cuda.Device(dev).use()
+                np = cupy
```
burlen, line 74: The metadata is CPU based. Unless we're doing a heavy calculation, probably best to use CPU in the report and request phase.

test/python/test_lapse_rate.py
```diff
@@ -146,7 +146,7 @@ def execute(self, port, data_in, req_in):
 
 
 def isequal(a, b, epsilon):
-    return np.fabs(a - b) < epsilon
+    return ((a - b)**2)**(0.5) < epsilon
```
burlen: You could use numpy's isclose or allclose for a more robust implementation. Note: TECA has C++ implementation in teca_coordinate_util::equal for use in C++ code.

burlen (commented):
@amandasd You got the cupy stuff squared away, but I think we shouldn't use the GPU on metadata in request and report methods. A couple other minor comments, mainly for your info. Sorry!

burlen:
@amandasd if this is ready, would you please rebase and merge?

PR #682, opened 2021-12-22
WIP -- implement the Numpy Array Interface protocol

burlen:


burlen:
This work was numpy has been superseded by #684, which took a different approach based on what was learned here.

PR #688, opened 2022-04-12
Temporal reduction

amandasd:
[WIP] C++ version of the temporal reduction application


alg/teca_temporal_reduction.cxx
```diff
@@ -0,0 +1,648 @@
...
+}
+
+// --------------------------------------------------------------------------
+void summation_operator::update(
+    p_teca_variant_array &out_array,
+    p_teca_variant_array &out_valid,
+    const_p_teca_variant_array &in_array,
+    const_p_teca_variant_array &in_valid)
+{
+    for (unsigned int i = 0; i < out_array->size(); ++i)
```
burlen, line 39: pull the call to `size` outside of the loop's condition. Function's in the condition get called before each iteration of the loop. There is more of a chance for compiler optimization when the condition is a simple comparison.

alg/teca_temporal_reduction.cxx
```diff
@@ -0,0 +1,648 @@
...
+    {
+
+        double array_in, array_out;
+        in_array->get(i, array_in);
+        out_array->get(i, array_out);
```
burlen, line 44: This is not the best API to use for heavy data processing.

1. This API fetches a single value. When the data is on the GPU this will pull a single value across the PCI bus in each iteration of the loop.
2. Data access API on `teca_variant_array` are virtual. There are 2 performance issues with calling virtual methods inside loops. a) virtual methods are resolved at runtime and as a result the compiler cannot optimize them. b) Dispatch (a big if/else that casts to the actual data type) is done in every loop iteration.
3. This API cannot be used on the GPU. It always puts data on the CPU.

Rather than use the virtual single value data access API defined in `teca_variant_array` we should:

1. Apply dispatch once outside of the loop. This will give us a typed variant array instance (ie `teca_variant_array_impl<T>`.  We use `TEMPLATE_DISPATCH` macro for this. 
2. Once you have the typed array, call the appropriate accessibility method. For eg if the load balancer assigns this thread to a CPU the device will be -1, and we'd call `teca_variant_array_impl<T>::get_cpu_accessible`. A pointer to the data then can be obtained, inside the loop you will use the pointer to access the data as needed.

Dispatch is done outside of the loop, once before the loop begins. Accessing the data inside the loop uses a pointer instead of a virtual call. The compiler can optimize this type of calculation.

You can use the `teca_l2_norm` as a template for this process.

burlen (commented):
I see a few things about how the loops are structured and how we are accessing the data, fixing these should improve the performance greatly.

alg/teca_temporal_reduction.cxx
```diff
@@ -0,0 +1,1083 @@
...
+
+    if (i >= n_elem)
+        return;
+
+//    if (p_in_valid == nullptr)
+//    {
+        if (p_in_array[i] < p_out_array[i])
+        {
+            p_out_array[i] = p_in_array[i];
+        }
```
burlen, line 66: this looks correct to me (no need for thrust here!)

alg/teca_temporal_reduction.cxx
```diff
@@ -0,0 +1,1083 @@
...
+        << streaming << " remain" << std::endl;
+#endif
+
+    // copy the first mesh
+    const_p_teca_cartesian_mesh mesh_in
+        = std::dynamic_pointer_cast<const teca_cartesian_mesh>(data_in[0]);
+    p_teca_cartesian_mesh mesh_out = teca_cartesian_mesh::New();
+    mesh_out->shallow_copy(
+        std::const_pointer_cast<teca_cartesian_mesh>(mesh_in));
+    p_teca_array_collection arrays_out = mesh_out->get_point_arrays();
```
burlen, line 1008: It's convenient to shallow copy the input mesh to grab metadata etc, but you'll need to allocate new arrays on the assigned device for the output. if you don't do this, a couple of bad things can happen. 1) you'll be modifying the input data, which is against the rules of the system 2) in the case the input was allocated on the CPU and you are trying to compute on the GPU you'll end up operating on a temporary and simply have the original (the one that was shallow copied) array at the end.

burlen, line 1008: It's convenient to shallow copy the input mesh to grab metadata etc, but you'll need to allocate new arrays on the assigned device for the output. if you don't do this, a couple of bad things can happen. 1) you'll be modifying the input data, which is against the rules of the system 2) in the case the input was allocated on the CPU and you are trying to compute on the GPU you'll end up operating on a temporary and simply have the original (the one that was shallow copied) array at the end.

alg/teca_temporal_reduction.cxx
```diff
@@ -0,0 +1,1083 @@
...
+#if defined(TECA_HAS_CUDA)
+    // set the CUDA device to run on
+    cudaError_t ierr = cudaSuccess;
+    if ((ierr = cudaSetDevice(device_id)) != cudaSuccess)
+    {
+        TECA_ERROR("Failed to set the CUDA device to " << device_id
+            << ". " << cudaGetErrorString(ierr))
+//        return -1;
+    }
+#endif
```
burlen, line 177: check for device_id >= 0 here

burlen, line 177: check for device_id >= 0 here

alg/teca_temporal_reduction.cxx
```diff
@@ -0,0 +1,1083 @@
...
+
+    unsigned long n_elem = out_array->size();
+
+    TEMPLATE_DISPATCH_FP_SI(
+        teca_variant_array_impl,
+        out_array.get(),
+
+#if defined(TECA_HAS_CUDA)
```
burlen, line 185: check for device_id >= 0

burlen, line 185: check for device_id >= 0

alg/teca_temporal_reduction.cxx
```diff
@@ -0,0 +1,1083 @@
...
+//
+//                    // update the valid value mask
+//                    p_out_valid[i] = 1;
+//                }
+//            }
+//#endif
+//        }
+//        else
+//        {
+#if defined(TECA_HAS_CUDA)
```
burlen, line 239: check for device_id >= 0

burlen, line 239: check for device_id >= 0

alg/teca_temporal_reduction.cxx
```diff
@@ -0,0 +1,1083 @@
...
+    {
+        TECA_ERROR("Failed to set the CUDA device to " << device_id
+            << ". " << cudaGetErrorString(ierr))
+//        return -1;
+    }
+#endif
+
+    unsigned long n_elem = out_array->size();
+
+    TEMPLATE_DISPATCH_FP_SI(
```
burlen, line 181: this version of the dispatch macro excludes unsigned integers, but do we need to exclude them here?

burlen, line 181: this version of the dispatch macro excludes unsigned integers, but do we need to exclude them here?

alg/teca_temporal_reduction.cxx
```diff
@@ -0,0 +1,1739 @@
...
+            p_teca_variant_array out_array = arrays_out->get(array);
+            if (arrays_out->has(valid))
+               out_valid = arrays_out->get(valid);
+
+            // finalize the reduction
+            p_teca_variant_array red_array;
+            this->op[array]->finalize(
+                device_id, out_array, out_valid, red_array);
+
+            // udpate the output
```
taobrienlbl, line 1723: udpate -> update

alg/teca_temporal_reduction.cxx
```diff
@@ -0,0 +1,1739 @@
...
+            }
+
+            // apply the reduction
+            p_teca_variant_array red_array;
+            p_teca_variant_array red_valid = nullptr;
+            this->op[array]->update(
+                device_id, out_array, out_valid,
+                in_array, in_valid, red_array, red_valid);
+
+            // udpate the output
```
taobrienlbl, line 1695: udpate -> update

apps/teca_temporal_reduction.cpp
```diff
@@ -0,0 +1,269 @@
...
+    mav->set_properties("mav", opt_vals);
+    cf_writer->set_properties("cf_writer", opt_vals);
+
+    bool have_file = opt_vals.count("input_file");
+    bool have_regex = opt_vals.count("input_regex");
+    if ((have_file && have_regex) || !(have_file || have_regex))
+    {
+        if (mpi_man.get_comm_rank() == 0)
+        {
+            TECA_FATAL_ERROR("Extacly one of --input_file or --input_regex can be specified. "
```
taobrienlbl, line 160: Extacly -> Exactly

data/teca_calendar_util.h
```diff
@@ -375,7 +413,7 @@ class TECA_EXPORT interval_iterator_factory
 public:
     /** Allocate and return an instance of the named iterator
      * @param[in] interval Name of the desired interval iterator. One of daily,
-     *                     monthly, seasonal, or yearly
+     *                     monthly, seasonal, yearly, or n_steps
```
taobrienlbl, line 416: Shouldn't line 422 include something like `n_steps = 6` in the enum?

taobrienlbl (commented):
Really nice job on this @amandasd !  I have very few substantial comments to give: mainly spelling errors.

The two substantial comments I do have are: (1) is there a unit test for the n_steps iterator, and (2) the documentation in `doc/rtd/applications.rst` should be updated with the new capabilities (i.e., the n_steps iterator, letting the user know that this is a GPU-enabled application, and anything else that differs from the python app).

Regarding point (1) above, I raise that because I noticed in `teca_calendar_util.h`, `n_steps` isn't included in an enum.  It looks to me like this enum isn't used in the code anywhere (not sure though?), but if it is, would it function properly without `n_steps` in the enum?

taobrienlbl:
Also, this will of course need to be rebased with develop.

In your last PR, it looked like you might have used `git merge develop` rather than `git rebase develop`; I think our general practice is to rebase branches when possible.  @burlen can clarify here.

taobrienlbl:
I'm testing out the `teca_temporal_reduction` app, and I have some additional comments.

* It looks like the default values indicated in the documentation aren't actually used.  For example, if I don't specify the --interval argument, the program errors out with a complaint that an interval hasn't been specified, even though the default appears to be 'monthly'
* I suggest using 'average' as the default reduction operator, as that will likely be the most common use case.
* I'm getting a segfault when I run it on a test dataset, but I'm not sure whether that's a problem with how I'm running it (or with the system I'm running on) or a bug in the code.  I'll follow up in slack about this.

alg/teca_temporal_reduction.h
```diff
@@ -0,0 +1,239 @@
...
+     */
+    ///@{
+    TECA_ALGORITHM_PROPERTY(double, fill_value)
+
+    /** @name use_fill_value
+     * Control how invalid or missing values are treated
+     * default 1
+     */
+    ///@{
+    TECA_ALGORITHM_PROPERTY(int, use_fill_value)
```
burlen, line 208: These can be removed. The missing value handling should all be automatic when used with the teca_valid_value mask. The Python temporal reduction predates the teca_valid_value_mask and these are no longer needed.

data/teca_calendar_util.h
burlen: Yes, it probably should. The issue with how I did this on the Python side was to get the number of steps as a string that could be parsed eg: "5_steps". That needs to change in order wo make use of the enum.

alg/teca_temporal_reduction.cxx
```diff
@@ -0,0 +1,1739 @@
...
+    try
+    {
+        t_atts.get("calendar", cal);
+    }
+    catch (...)
+    {
+        cal = "standard";
+        TECA_WARNING("Attributes for the time axis " << t_var <<
+            " is missing calendar. The \"standard\" calendar will be used")
+    }
```
burlen, line 1414: In C++ no exception will be thrown. Instead test for the presence of the key.

alg/teca_temporal_reduction.cxx
```diff
@@ -0,0 +1,1739 @@
...
+            }
+        }
+
+        // create and initialize the operator
+        p_reduction_operator op
+                      = reduction_operator_factory::New(this->operator_name);
+        op->initialize(fill_value);
+
+        // save the operator
+        this->op[array] = op;
```
burlen, line 1597: this needs to be thread safe, and thread local.

alg/teca_temporal_reduction.cxx
```diff
@@ -0,0 +1,1739 @@
...
+      std::const_pointer_cast<teca_cartesian_mesh>(mesh_in));
+    p_teca_array_collection arrays_out = mesh_out->get_point_arrays();
+
+
+    // accumulate incoming values
+    for (int i = n_data-2; i >= 0; --i)
+    {
+        mesh_in
+           = std::dynamic_pointer_cast<const teca_cartesian_mesh>(data_in[i]);
+        const_p_teca_array_collection arrays_in = mesh_in->get_point_arrays();
```
burlen, line 1663: I find it counter intuitive as to why we are processing the list in reverse order.  I suspect it is a hold over of the Python implementation which pop'd elements off the list. I don't think the order matters here. It's probably best to process in ascending order, just because thats what we always do. If not, comment the heck out of it.

Also, pull calculations of the loop termination condition.

alg/teca_temporal_reduction.cxx
```diff
@@ -0,0 +1,1739 @@
...
+            {
+               out_valid = arrays_out->get(valid);
+            }
+
+            // apply the reduction
+            p_teca_variant_array red_array;
+            p_teca_variant_array red_valid = nullptr;
+            this->op[array]->update(
+                device_id, out_array, out_valid,
+                in_array, in_valid, red_array, red_valid);
```
burlen, line 1693: this needs to be a thread safe and thread local storage

apps/teca_temporal_reduction.cpp
```diff
@@ -0,0 +1,269 @@
...
+    else
+    {
+        mav->set_use_fill_value(0);
+    }
+
+    if (opt_vals.count("point_arrays"))
+        mav->set_point_arrays(opt_vals["point_arrays"].as<std::vector<std::string>>());
+
+    cf_writer->set_input_connection(mav->get_output_port());
+    cf_writer->set_thread_pool_size(1);
```
burlen, line 236: This should be -1 and there should be a command line argument to modify it

apps/teca_temporal_reduction.cpp
```diff
@@ -0,0 +1,269 @@
...
+       std::cerr << "point_arrays=" << opt_vals["point_arrays"].as<std::vector<std::string>>()<< std::endl;
+       std::cerr << "ignore_fill_value=" << opt_vals["ignore_fill_value"].as<int>() << std::endl;
+       std::cerr << "fill_value=" << opt_vals["fill_value"].as<double>() << std::endl;
+    }
+
+    mcf_reader->set_properties("mcf_reader", opt_vals);
+    cf_reader->set_properties("cf_reader", opt_vals);
+    vv_mask->set_properties("vv_mask", opt_vals);
+    unpack->set_properties("unpack", opt_vals);
+    mav->set_properties("mav", opt_vals);
```
burlen, line 151: Use a more descriptive name here because this is displayed in the help message and used to set advanced property values.

burlen (commented):
Hi Amanda, You've done a ton of work. I love it, great job. I found a few things.  The main one is thread safety issues. Python code is single threaded but the C++ code is multi-threaded. The per-array operators need to be held in thread local storage and access to the container needs to be thread safe.

alg/teca_temporal_reduction.cxx
```diff
@@ -0,0 +1,1739 @@
...
+            std::string &array = this->point_arrays[j];
+
+            // the valid value masks
+            std::string valid = array + "_valid";
+            const_p_teca_variant_array in_valid = nullptr;
+            const_p_teca_variant_array out_valid = nullptr;
+
+            // arrays
+            const_p_teca_variant_array in_array = arrays_in->get(array);
+            const_p_teca_variant_array out_array = arrays_out->get(array);
```
burlen, line 1676: check for nullptr an report an error if the array is not found

taobrienlbl (approved):
Aside from the outstanding issues we have been discussing in e-mail (e.g., rebasing, yet another difficult dataset, etc), this looks good to go from my perspective!

burlen:
@amandasd @taobrienlbl This is in much better shape now! However, I think this is going to need some revision and cleanup before merging. One of the main issue now is the design relating to threading. I will make a review shortly to provide specific details.

burlen (changes requested):
Hi Amanda, You done a ton of work on this and I'm pleased with your progress.  This is one of our larger and complex pipeline stages. Direct port from Python to C++ is not always possible or the best way because of the differences between languages. The major part of the work that remains is resolving those type of issues. I think we need to change the API for reduction operators. Let's talk more about that next week. In some cases I asked for changes to code that was used in Python that does port cleanly but was leftover from older revisions and needed cleanup. A few items from the last review slipped through the cracks. I resolved the issues in the last review I thought were fixed, so please look at the unresolved conversations and see if those can be addressed too. Let me know if you have questions about any of the comments. Thanks for your hard work on this!

alg/teca_temporal_reduction.cxx
```diff
@@ -0,0 +1,2147 @@
...
+
+    using time_interval_t = time_interval;
+    using p_reduction_operator = std::shared_ptr<reduction_operator_collection>;
+
+public:
+    std::vector<time_interval_t> indices;
+    std::map<std::string, p_reduction_operator> op;
+};
+
+class teca_temporal_reduction::internals_t::reduction_operator_collection
```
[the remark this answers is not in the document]
taobrienlbl: and to give an idea of custom reductions that one might want to do (that we might want to implement later on): standard deviation and percentile (e.g., what's the 95th percentile value for each grid cell)

amandasd:
Thanks @burlen and @taobrienlbl for your review comments. I will work on them!

apps/teca_temporal_reduction.cpp
amandasd: Do we need the two below or only one of them?

    if (!opt_vals["n_threads"].defaulted())
        mav->set_thread_pool_size(opt_vals["n_threads"].as<int>());
    else
        mav->set_thread_pool_size(-1);

    if (!opt_vals["n_threads"].defaulted())
        cf_writer->set_thread_pool_size(opt_vals["n_threads"].as<int>());
    else
        cf_writer->set_thread_pool_size(-1);

apps/teca_temporal_reduction.cpp
```diff
@@ -0,0 +1,269 @@
...
+    }
+
+    unpack->set_input_connection(vv_mask->get_output_port());
+
+    mav->set_input_connection(unpack->get_output_port());
+    mav->set_stream_size(2);
+
+    if (!opt_vals["n_threads"].defaulted())
+    {
+        mav->set_thread_pool_size(opt_vals["n_threads"].as<int>());
```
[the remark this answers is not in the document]
amandasd: I got segmentation fault when setting the thread_pool as below. The n_threads default value is -1 and we are not setting any value for n_threads. Am I missing something?

mav->set_thread_pool_size(1);

if (!opt_vals["n_threads"].defaulted())
{   
    cf_writer->set_thread_pool_size(opt_vals["n_threads"].as<int>());
}   
else
{   
    cf_writer->set_thread_pool_size(-1);
}

alg/teca_temporal_reduction.h
amandasd: @burlen Should I remove only line 208 or lines 201 and 208?

apps/teca_temporal_reduction.cpp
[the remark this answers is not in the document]
burlen: can you share a stack trace?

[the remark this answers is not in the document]
amandasd: Here is the strace output:

```
...
pwrite64(31, "\0\0\0\0\220\325\341@\0\0\0\0p\331\341@\0\0\0\0\360\334\341@\0\0\0\0\320\340\341@"..., 96, 2048) = 96
pwrite64(31, "\267\356\346\251\16IV\300]\341].\342\1V\300c\vA\16J\272U\300\"O\222\256\231rU\300"..., 1280, 2144) = 1280
pwrite64(31, "\0\0\0\0\0\0\0\0\0\0\0\0\0\0\362?\0\0\0\0\0\0\2@\0\0\0\0\0\0\v@"..., 2560, 6144) = 2560
clock_nanosleep(CLOCK_REALTIME, 0, {tv_sec=0, tv_nsec=1000000}, 0x7ffde48c8f80) = 0
clock_nanosleep(CLOCK_REALTIME, 0, {tv_sec=0, tv_nsec=1000000}, 0x7ffde48c8f80) = 0
clock_nanosleep(CLOCK_REALTIME, 0, {tv_sec=0, tv_nsec=1000000}, 0x7ffde48c8f80) = 0
clock_nanosleep(CLOCK_REALTIME, 0, {tv_sec=0, tv_nsec=1000000}, [0:140383497191424]teca_temporal_reduction::get_upstream_request
[0:140383765757952]teca_temporal_reduction::get_upstream_request
[0:140383782543360]teca_temporal_reduction::get_upstream_request
[0:140384017539072]teca_temporal_reduction::get_upstream_request
[0:140383278981120]teca_temporal_reduction::get_upstream_request
[0:140383623081984]teca_temporal_reduction::get_upstream_request
[0:140383438442496][0:140383631474688]teca_temporal_reduction::get_upstream_request
teca_temporal_reduction::get_upstream_request
[0:140383614689280]teca_temporal_reduction::get_upstream_request
[0:140383707009024]teca_temporal_reduction::get_upstream_request
0x7ffde48c8f80) = 0
clock_nanosleep(CLOCK_REALTIME, 0, {tv_sec=0, tv_nsec=1000000}, [0:140383320944640]teca_temporal_reduction::get_upstream_request
[0:140383354515456]teca_temporal_reduction::get_upstream_request
0x7ffde48c8f80) = 0
clock_nanosleep(CLOCK_REALTIME, 0, {tv_sec=0, tv_nsec=1000000}, 0x7ffde48c8f80) = 0
clock_nanosleep(CLOCK_REALTIME, 0, {tv_sec=0, tv_nsec=1000000},  <unfinished ...>) = ?
+++ killed by SIGSEGV +++
Segmentation fault
```

amandasd:
The datasets that are triggering the segfault is at `/global/cfs/cdirs/m1517/cascade/taobrien/ARTMIP-Tier2-Paleo/standardized_dataset/`. 
In case you want to take a look, please find attached the gdb output for this case.
[gdb-dataset-cpu.txt](https://github.com/LBL-EESA/TECA/files/9144719/gdb-dataset-cpu.txt)
[gdb-dataset-gpu.txt](https://github.com/LBL-EESA/TECA/files/9144720/gdb-dataset-gpu.txt)

data/teca_calendar_util.h
```diff
@@ -167,6 +167,8 @@ class TECA_EXPORT interval_iterator
     virtual int get_next_interval(time_point &first_step,
         time_point &last_step) = 0;
 
+    virtual int set_number_of_steps(long n) {}
```
burlen, line 170: this is specific to the n_steps iterator and therefor should not be a virtual in the base class because the base class is for api common to all the iterators.

data/teca_calendar_util.cxx
```diff
@@ -783,6 +785,92 @@ int day_iterator::get_next_interval(time_point &first_step,
...
+    {
+        return false;
+    }
+
+    return true;
+}
+
+int n_steps_iterator::set_number_of_steps(long n)
+{
+    this->number_of_steps = n;
```
burlen, line 804: missing return statement

data/teca_calendar_util.cxx
```diff
@@ -783,6 +785,92 @@ int day_iterator::get_next_interval(time_point &first_step,
...
+}
+
+// --------------------------------------------------------------------------
+int n_steps_iterator::initialize(const const_p_teca_variant_array &t,
+    const std::string &units, const std::string &calendar,
```
burlen, line 809: calendar and units are not used in this function so cast to void to silence the compiler warning. eg `(void) calendar;`

alg/teca_temporal_reduction.cxx
```diff
@@ -0,0 +1,2324 @@
...
+     *  finalize:
+     *      finalizes arrays_out (current state) and returns the result.
+     *      If no finalization is needed simply return arrays_out.
+     */
+
+    reduction_operator() : fill_value(-1) {}
+
+    virtual void initialize(double fill_value);
+
+    virtual int update_cpu(int device_id,
```
burlen, line 78: there are a bunch of compiler warnings relating to these functions, eg not return statement.  I think you could make these pure virtual since it would be a problem if the derived classes did not implement them.

alg/teca_temporal_reduction.cxx
```diff
@@ -0,0 +1,2324 @@
...
+            if (out_valid != nullptr)
+            {
+                auto sp_count = static_cast<TT*>(count.get())->get_cuda_accessible();
+                NT *p_count = sp_count.get();
+
+                auto sp_out_valid = static_cast<const TT_MASK*>(out_valid.get())->get_cuda_accessible();
+                const NT_MASK *p_out_valid = sp_out_valid.get();
+
+#if defined(TECA_HAS_CUDA)
+                cuda::count_init(device_id, p_out_valid, p_count, n_elem);
```
burlen, line 932: missing cpu case here

burlen (commented):
Hi Amanda, I started to look into the threading issues this morning. I'm still working on this. the compiler is throwing lots of warnings for the CPU only build  I put a few comment just now after looking into them. Would you please clean up all the warnings when you get a chance?

burlen:
@amandasd I made the threading fixes and pushed them here. all seems good. ready for you.

unrelated : 2 tests fail in the cuda build. please take a look at the following tests. note that the Travis-CI dahsboards don't run CUDA build so we won't see them there:

	163 - test_temporal_reduction_app_daily_maximum_missing_values (Failed)
	164 - test_temporal_reduction_app_daily_minimum_missing_values (Failed)

burlen:
the file name of the original python implementation was changed but not the class name. this is resulting in 2 classes of the same name (teca_temporal_reduction) the second one wins in this case the old implementation because it comes second in the .i.

the class name  of the original python implementation needs to be changed. I think all of this renaming of the old class belongs in its own commit.

the original tests and app needs to keep working for the time being so those will also need to be renamed.

test/apps/CMakeLists.txt
```diff
@@ -285,48 +304,48 @@ teca_add_app_test(test_temporal_reduction_app_monthly_maximum
     COMMAND ${CMAKE_CURRENT_SOURCE_DIR}/test_temporal_reduction_app.sh
     ${CMAKE_BINARY_DIR}/${BIN_PREFIX} ${TECA_DATA_ROOT}
     "prw_hus_day_MRI-CGCM3_historical_r1i1p1_19500101-19501231\\.nc" prw
-    monthly maximum 7 0
+    monthly maximum -1 7 0
```
amandasd, line 307: @burlen wouldn't it be `monthly maximum 7 -1 0` what you mean?

burlen: yes, sorry about that one!

amandasd: No problem, I will fix it.

burlen: Thank you, sorry for the trouble

burlen:
> The datasets that are triggering the segfault is at `/global/cfs/cdirs/m1517/cascade/taobrien/ARTMIP-Tier2-Paleo/standardized_dataset/Guan_Waliser_v2/PreIndust`. T
> 
> `srun -N 1 -n 1 bin/teca_temporal_reduction --input_regex '../../PreIndust/PreIndust.ar_tag.teca-bard-v1.0.1.6hr.015.*\.nc' --interval=monthly --point_arrays ar_binary_tag --operator=average --file_layout yearly --output_file PreIndust.ar_tag.teca-bard-v1.0.1.monthly.%t%.nc --verbose 1`

@amandasd Have you ruled out the data type "byte" being an issue?
```
	byte ar_binary_tag(time, lat, lon) ;
		ar_binary_tag:long_name = "Binary Indicator of Atmospheric River" ;
		ar_binary_tag:units = "none" ;
		ar_binary_tag:description = "binary indicator of atmospheric river" ;
		ar_binary_tag:scheme = "Guan&Waliser" ;
		ar_binary_tag:version = "2" ;
```
To me that seems to be a likely candidate and good place to start

amandasd:
> > The datasets that are triggering the segfault is at `/global/cfs/cdirs/m1517/cascade/taobrien/ARTMIP-Tier2-Paleo/standardized_dataset/Guan_Waliser_v2/PreIndust`. T
> > `srun -N 1 -n 1 bin/teca_temporal_reduction --input_regex '../../PreIndust/PreIndust.ar_tag.teca-bard-v1.0.1.6hr.015.*\.nc' --interval=monthly --point_arrays ar_binary_tag --operator=average --file_layout yearly --output_file PreIndust.ar_tag.teca-bard-v1.0.1.monthly.%t%.nc --verbose 1`
> 
> @amandasd Have you ruled out the data type "byte" being an issue?
> 
> ```
> 	byte ar_binary_tag(time, lat, lon) ;
> 		ar_binary_tag:long_name = "Binary Indicator of Atmospheric River" ;
> 		ar_binary_tag:units = "none" ;
> 		ar_binary_tag:description = "binary indicator of atmospheric river" ;
> 		ar_binary_tag:scheme = "Guan&Waliser" ;
> 		ar_binary_tag:version = "2" ;
> ```
> 
> To me that seems to be a likely candidate and good place to start

@burlen I don't think so. It works well if you run without the last file (the last file has only one time; all other files have 1460 times, 4 times per day).

PR #691, opened 2022-05-10
Spatial parallel

burlen:
extend the system to process the time axis by partitioning over space


burlen:
spatial_partitioning
Spatial partitioning, 8 ranks, 768 time steps, temporal block size 12



space_time_partitioning
Space-time partitioning: 10 ranks, 768 time steps, temporal block size 12, 16 spatial partitions

The color shows the MPI rank that a given partition was assigned to

burlen:
era5_seaosnal_avg_with_spatial_paralelelism

some runs on 41 years of ERA5 on Cori with space-time partitioning enabled. The first data point is temporal only max parallelism. Spatial partitions are used to increase the number of MPI ranks in use. Scaling is less than ideal likely due to increased I/O overheads.

PR #694, opened 2022-05-17
Adding a gradient algorithm

taobrienlbl:


burlen (commented):
Hi Travis, The calculation looks correct for a uniform Cartesian mesh.  I found a few metadata related issues that should be fixed.

alg/teca_gradient.cxx
```diff
@@ -0,0 +1,406 @@
...
+
+// --------------------------------------------------------------------------
+std::string teca_gradient::get_gradient_field_y(
+    const teca_metadata &request)
+{
+    std::string grad_y_var = this->gradient_field_y;
+
+    if (grad_y_var.empty() &&
+        request.has("teca_gradient::gradient_field_y"))
+            request.get("teca_gradient::gradient_field_y", grad_y_var);
```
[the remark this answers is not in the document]
taobrienlbl: done!

alg/teca_gradient.cxx
```diff
@@ -0,0 +1,406 @@
...
+
+// --------------------------------------------------------------------------
+std::string teca_gradient::get_gradient_field_x(
+    const teca_metadata &request)
+{
+    std::string grad_x_var = this->gradient_field_x;
+
+    if (grad_x_var.empty() &&
+        request.has("teca_gradient::gradient_field_x"))
+            request.get("teca_gradient::gradient_field_y", grad_x_var);
```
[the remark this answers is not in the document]
taobrienlbl: 😬

alg/teca_gradient.cxx
```diff
@@ -0,0 +1,406 @@
...
+#ifdef TECA_DEBUG
+    cerr << teca_parallel_id()
+        << "teca_gradient::get_output_metadata" << endl;
+#endif
+    (void)port;
+
+    // add in the arrays we will generate
+    teca_metadata out_md(input_md[0]);
+    out_md.append("variables", this->gradient_field_x);
+    out_md.append("variables", this->gradient_field_y);
```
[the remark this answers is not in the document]
taobrienlbl: done!

alg/teca_gradient.cxx
```diff
@@ -0,0 +1,406 @@
...
+    const std::vector<teca_metadata> &input_md)
+{
+#ifdef TECA_DEBUG
+    cerr << teca_parallel_id()
+        << "teca_gradient::get_output_metadata" << endl;
+#endif
+    (void)port;
+
+    // add in the arrays we will generate
+    teca_metadata out_md(input_md[0]);
```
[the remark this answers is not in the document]
taobrienlbl: I see -- that makes sense.  I fixed this and updated the SVN repo.

As a result, the data file referenced in #695 no longer exists, though one could grab it from commit 138 (or I could regenerate it from the script I used).

taobrienlbl:
I believe I've addressed all your comments @burlen

@amandasd - do you have any?

alg/teca_gradient.cxx
```diff
@@ -0,0 +1,464 @@
...
+            grad_x[j*n_lon + max_i] = num_t();
+            grad_y[j*n_lon + max_i] = num_t();
+        }
+    }
+
+    // extend y-derivative values into lat boundaries
+    num_t *dest = grad_y;
+    num_t *src = grad_y + n_lon;
+    for (unsigned long i = 0; i < n_lon; ++i)
+        dest[i] = src[i+n_lon];
```
amandasd, line 130: Wouldn't it be dest[i] = src[i] in line 130?

alg/teca_gradient.cxx
```diff
@@ -0,0 +1,464 @@
...
+            const num_t *f_y2 = f + jj + n_lon;
+            const num_t *f_y0 = f + jj - n_lon;
+            const num_t *f_x2 = f + jj + 1;
+            const num_t *f_x0 = f + jj + max_i;
+            num_t *gx = grad_x + jj;
+            num_t *gy = grad_y + jj;
+            num_t dx = delta_x[j];
+
+            gx[0] = (f_x2[0] - f_x0[0]) / dx;
+            gy[0] = (f_y2[0] - f_y0[0]) / dx;
```
amandasd, line 92: Wouldn't it be dy instead of dx in line 92?

taobrienlbl: yes, thank you on this and the one below!!  The test wouldn't have caught this because dx=dy

alg/teca_gradient.cxx
taobrienlbl: I see what you're saying.  It does look like that, but I'm confused because this code reproduced a python version of the same boundary condition (implemented totally differently, so it's unlikely the same mistake was made twice).

@burlen, this is based on `teca_vorticity.cxx` (see [line 117](https://github.com/LBL-EESA/TECA/blob/83512cb29765d1f578a2615a4a91e0262af5b446/alg/teca_vorticity.cxx#L117)), so if there's a problem here, there is a problem there too.  I always get turned around when thinking about pointer math though.  What do you think?

alg/teca_gradient.cxx
```diff
@@ -0,0 +1,464 @@
...
+// compute the gradient using second-order, centered finite difference
+// assumes fixed mesh spacing. here we add periodic bc in lon
+// and apply unit stride vector optimization strategy to loops
+template <typename num_t, typename pt_t>
+void gradient(num_t *grad_x, num_t *grad_y, const pt_t *lon, const pt_t *lat,
+    const num_t *f, unsigned long n_lon,
+    unsigned long n_lat, bool periodic_lon=true)
+{
+    size_t n_bytes = n_lat*sizeof(num_t);
+    num_t *delta_x = static_cast<num_t*>(malloc(n_bytes));
```
amandasd, line 44: Isn't better to use std::vector<num_t> instead of explicitly malloc?

taobrienlbl: That would be more in the C++ style, yes.  I copied the approach from `teca_vorticity.cxx`, so I'm wondering if there's a performance reason for using a dynamically allocated array rather than a vector: perhaps the overhead from bounds-checking in a vector?  @burlen  - do you have an opinion here on which approach should be used?

taobrienlbl: I ended up going with `std::vector` since the `[]` operator doesn't have any overheads.

alg/teca_gradient.cxx
```diff
@@ -0,0 +1,464 @@
...
+            const num_t *f_y2 = f + jj + max_i + n_lon;
+            const num_t *f_y0 = f + jj + max_i - n_lon;
+            const num_t *f_x2 = f + jj;
+            const num_t *f_x0 = f + jj + max_i - 1;
+            num_t *gx = grad_x + jj + max_i;
+            num_t *gy = grad_y + jj + max_i;
+            num_t dx = delta_x[j];
+
+            gx[0] = (f_x2[0] - f_x0[0]) / dx;
+            gy[0] = (f_y2[0] - f_y0[0]) / dx;
```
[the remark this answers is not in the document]
taobrienlbl: this has been fixed, thanks!

alg/teca_gradient.cxx
taobrienlbl: this has been fixed, thanks!

alg/teca_gradient.cxx
taobrienlbl: I can confirm that this was a pointer math bug. (And it turns out there was a bug in the code I used to compare with the  python-generated version, such that I wasn't actually comparing TECA's gradient at all.)

I've also submitted a fix to `teca_vorticity.cxx`.

burlen: @taobrienlbl sorry for late reply, agree. Good catch @amandasd!

alg/teca_gradient.cxx
taobrienlbl: OK, I'll revert it.

burlen: @taobrienlbl It's fine  to use vector here, I didn't mean for you to have to revert this, just wanted to give an answer as to why I use malloc.

taobrienlbl:
Oh I see, thanks for clarifying. It’s been reverted in any case.

taobrienlbl:
@burlen - when and if you think this is ready, would you please approve for merging?

PR #708, opened 2022-07-25
Temporal index select

taobrienlbl:
Wrote teca_temporal_index_select: an algorithm for selecting specific time indices. Added as an optional stage in teca_temporal_reduction.

(Note: this is a repeat of #706, with the commits squashed.)


alg/teca_temporal_index_select.cxx
```diff
@@ -0,0 +1,337 @@
...
+    // default to CPU operations
+    int device_id = -1;
+    teca_variant_array::allocator allocator = 
+        teca_variant_array::allocator::malloc;
+#if defined(TECA_HAS_CUDA)
+    request.get("device_id",device_id);
+
+    // set the CUDA device to run on
+    cudaError_t ierr = cudaSuccess;
+    if (( ierr = cudaSetDevice(device_id)) != cudaSuccess )
```
burlen, line 238: put this in the `if (device_id >= 0)` block.

burlen (commented):
I see one thing: don't set the cuda device if we are assigned a CPU (device_id == -1) this is not a valid cuda device id.

alg/teca_temporal_index_select.cxx
taobrienlbl: good catch, thanks; done

taobrienlbl:
@burlen - would you mind giving a final check and approval when you get a moment, please?

burlen (approved):
Awesome! Thanks Travis

PR #710, opened 2022-07-27
Spectral filter

burlen:
the spectral filter processes the time domain and makes use of spatial parallelism. this work includes I/O bench marking and optimizations that make use of MPI collective buffering in HDF5. An important factor in leveraging these new optimizations is to set the stripe count on the input/output directories. This is currently manual (the user has to do this in the shell before the run).  Striping on the input dir will always be manual, but striping on the output dir could be automated. I will leave this for a future PR.


burlen:
## Low pass filtering regression test
test_spectral_filter_low_pass


## High pass filtering regression test
test_spectral_filter_high_pass

Note: these have 1000 samples for 2 years of data. To resolve the daily signal would need a lot more samples.

burlen:
@taobrienlbl would you mind taking a look at the `teca_spectral_filter` code and let me know if you see any issues?

alg/teca_spectral_filter.py
```diff
@@ -0,0 +1,253 @@
...
+            return 60.
+        elif 'hours' in  unit:
+            return 60.*60.
+        elif 'days' in unit:
+            return 60.*60.*24.
+
+        raise ValueError('Unsupported time axis units %s' % (unit))
+
+    @staticmethod
+    def butterworth(npmod, n_samples, sample_rate, critical_freq, order, high_pass):
```
taobrienlbl, line 86: I suggest not trying to implement the butterworth filter directly, rather take advantage of its implementation in scipy:  https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.butter.html

test/python/test_spectral_filter.py
```diff
@@ -0,0 +1,331 @@
...
+
+        i = 0
+        while i < nt:
+            t = t0 + i*dt
+            nf = len(self.frequencies)
+            j = 0
+            while j < nf:
+                f = self.frequencies[j]
+                A = self.amplitudes[j]
+                f_t[i,:,:,:] += A * npmod.sin( 2*pi*f*t )
```
taobrienlbl, line 153: It might be a good idea to make a test based on the example given in the [scipy.signal.butter documentation](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.butter.html), as that is a very simple test case with an easy-to-understand result (only the 20 hz wave should survive).

alg/teca_spectral_filter.py
```diff
@@ -0,0 +1,253 @@
...
+                array_in = arrays_in[array_name_in].get_cpu_accessible()
+            else:
+                array_in = arrays_in[array_name_in].get_cuda_accessible()
+
+            # reshape the input array
+            nx,ny,nz,nt = mesh_in.get_array_shape(array_name_in)
+            array_in.shape = (nt,nz,ny,nx)
+
+            # apply the filter in the frequency domain
+            array_out = npmod.fft.ifft( npmod.fft.ifftshift(
```
taobrienlbl, line 230: I think the application of the butterworth filter is more complicated than just a simple fft-based filter.  It's a type of filter I'm not too familiar with, though, so I can't comment more.  But to get an idea of the complexity, look at the source code for [scipy.signal.butter](https://github.com/scipy/scipy/blob/v1.8.1/scipy/signal/_filter_design.py#L2935-L3046) and [scipy.signal.sosfilt](https://github.com/scipy/scipy/blob/v1.8.1/scipy/signal/_signaltools.py#L4129-L4238), which are the two routines that would be used to apply the butterworth filter using scipy.

alg/teca_spectral_filter.py
```diff
@@ -0,0 +1,253 @@
...
+        n_samples - filter size in number of samples
+        sample_rate - sample rate in Hz
+        critical_freq - cut off frequency in Hz
+        order - filter order
+        high_pass - if true the high pass version is generated
+        """
+        fs2 = sample_rate / 2.
+        f = npmod.linspace(-fs2, fs2, n_samples)
+        if high_pass:
+            h =  1. - 1. / npmod.sqrt( 1. + npmod.power( f / critical_freq, 2*order ) )
```
taobrienlbl, line 101: Only the zero-frequency part of the high-pass has 1 added to it.  Because of this, converting from a high-pass to a low-pass filter should take two lines, something like:
```python
h_high = -h_low
h_high[0] += 1
```

alg/teca_spectral_filter.py
taobrienlbl: See also [DSP Guide Chapter 20](http://www.dspguide.com/ch20/2.htm) for a discussion of this type of filter (which differs from a convolutional filter).

test/python/test_spectral_filter.py
burlen: The regression test does exactly this although the defaults test in the parameter space we are interested in. For instance here is how to reproduce the numpy example:
```
$ python3 "/home/bloring/work/teca/teca/test/python/test_spectral_filter.py" "90" "45" "2" "1000" "1" "1" "1" "0" "0.1" "0.05" "0.6667" "seconds" "high_pass" "analytic" "/home/bloring/work/teca/TECA_data/test_spectral_filter_low_pass" "1"

$ display test_spectral_filter_high_pass.png
```

And here is the plot the test generated:
test_spectral_filter_high_pass

alg/teca_spectral_filter.py
burlen: I suppose I could have, however I found it overly complicated and wanted something that would easily translate to C++. Having had the experience of writing the simple version myself, I now feel more comfortable dealing with scpipy's complexity.

alg/teca_spectral_filter.py
burlen: What I'm doing is described in chapter 18 of the book your pointing to. What you're pointing to here in chap 20 is about chebychev filers and uses a different approach - the z-transform.

alg/teca_spectral_filter.py
burlen: I took 1 and subtracted the low pass response which then "inverts" the response. Here's a plot for the numpy example you referenced w/ critical frequency 15 hz:
bw_high_pass
You can see the low frequencies are filtered while the high frequencies are passed. I think you're saying I should pass the "DC frequency" as well when doing a high pass filter. If so I think that make sense.

alg/teca_spectral_filter.py
burlen: I think this is worth looking into these methods, although it's going to take more time than I have before my talk. I think I will certainly emphasize that what I'm doing is not ideal and is only for demonstration purposes.

alg/teca_spectral_filter.py
burlen: this may be an easy change, I'll look into it.

test/python/test_spectral_filter.py
burlen: note: you can see from the filter's frequency response in this case  (image in a comment below) that 10Hz signal is not going to be completely filtered. I think this explains the error in the 20 Hz wave amplitude here.

taobrienlbl: Cool!  I think this raises an important question though: why does the solution at the bottom of the [scipy.signal.butterworth](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.butter.html) documentation have no evidence of the 10 Hz signal?  It could be because your example uses a 2nd order filter (if I read the code correctly) and theirs uses a 10th-order filter.

alg/teca_spectral_filter.py
taobrienlbl: > What I'm doing is described in chapter 18
Yes, I understand that.  The point I was making was that I don't think the butterworth filter is traditionally applied via the convolution theorem, and I think there's a good reason for that (but one that I don't have direct experience with).

taobrienlbl: Also, if I read the code correctly, it looks like the FFT is applied to the data directly without any zero-padding and tapering.  The use of the FFT assumes that the data are periodic--which is true in the test cases that you show. But in real-world data it isn't, and direct application of FFT-based transforms without zero-padding will result data from the end of the timeseries bleeding onto data at the beginning of the timeseries (and vice versa).  Tapering of some kind is necessary when zero-padding because otherwise you're effectively multiplying by a top-hat function in the temporal domain which leads to severe ringing in the spectral domain, which messes up the filter characteristics.

I suggest adding a test that isn't exactly periodic.

alg/teca_spectral_filter.py
taobrienlbl: I think that might work in this specific case, but it can be proved (I have this proof in a notebook somewhere, and I'll upload it if I track it down) that the formulation that I describe above is exactly identical to taking the original signal and subtracting-off the low-pass filter, hence leaving the high-frequency components.

burlen: OK. Got it. I also found an explicit formula for the Butterworth high pass [here]( https://tttapa.github.io/Pages/Mathematics/Systems-and-Control-Theory/Analog-Filters/Butterworth-Filters.html) and have updated to use that. here's the order 10 from the updated example above:
butter_kernel_high pass

test/python/test_spectral_filter.py
burlen: Yes, You're right, I used order 2. When I use order 10 I get a better result.
test_spectral_filter_high_pass

alg/teca_spectral_filter.py
burlen: OK. You're right. tried to test that, but now that I think about it all my window sizes were exact periods of the sin waves.

taobrienlbl:
@burlen I can't think of any further comments.  Nice work on this!

alg/teca_spectral_filter.py
burlen: do you know if I use the scipy routines if I'll

> @burlen I can't think of any further comments. Nice work on this!

@taobrienlbl Thanks. I'm having some doubts about weather or not this should be used/presented in this form. I could use scipy signal instead but I do not think that will work on the GPU, and I think I'll still have to deal with the windowing issues.I might be able to run with one window.

burlen:
@taobrienlbl I tried using scipy signal butter and sos filtering as this was an easy change to make. there is really bad distortion at the start of the filtered signal. Here is the run of their example, I reproduce their result:
test_spectral_filter_high_pass_0-999
Notice the errors on the left side of the filtered signal. The MSE for the sos based method is 0.52 and for my fft based method 0.0002

When I try this on the yearly "temperature" case that I'm using to verify climate scale units it's also worse.
test_spectral_filter_low_pass_0-999

MSE for sosfilt is 650 and for my fft based method 0.00157208

any thoughts on why the signal is so distorted at the left side using scipy signal? and if that type of error would be acceptable in practice?

for reference here's a diff of the changes I made to test scipy signal:

```diff
smic:~/work/teca/teca$git diff HEAD..stash@{0}
diff --git a/alg/teca_spectral_filter.py b/alg/teca_spectral_filter.py
index 57ced671..55a7f25d 100644
--- a/alg/teca_spectral_filter.py
+++ b/alg/teca_spectral_filter.py
@@ -1,6 +1,7 @@
 import sys, time
 import argparse
 import numpy
+from scipy import signal
 if get_teca_has_cuda() and get_teca_has_cupy():
     import cupy
 
@@ -204,14 +205,17 @@ class teca_spectral_filter(teca_python_algorithm):
         sample_rate = win_size / ( t_win * seconds_per )
 
         # generate the filter kernel
-        high_pass = 1 if self.filter_type == 'high_pass' else 0
+        #high_pass = 1 if self.filter_type == 'high_pass' else 0
 
-        H = self.butterworth(npmod, win_size, sample_rate,
-                        self.critical_frequency, self.filter_order,
-                        high_pass)
+        #H = self.butterworth(npmod, win_size, sample_rate,
+        #                self.critical_frequency, self.filter_order,
+        #                high_pass)
 
         # reshape so that numpy broadcasting works
-        H.shape = [win_size, 1, 1, 1]
+        #H.shape = [win_size, 1, 1, 1]
+
+        sos = signal.butter(self.filter_order, self.critical_frequency,
+            btype=self.filter_type.replace('_',''), fs=sample_rate, output='sos')
 
         # apply the filter in the frequenmcy domain
         arrays_in = mesh_in.get_point_arrays()
@@ -230,10 +234,12 @@ class teca_spectral_filter(teca_python_algorithm):
             array_in.shape = (nt,nz,ny,nx)
 
             # apply the filter in the frequency domain
-            array_out = npmod.fft.ifft( npmod.fft.ifftshift(
-                            npmod.fft.fftshift( npmod.fft.fft(
-                                array_in , axis=0 ), axes=0) * H,
-                                    axes=0 ), axis=0 )
+            #array_out = npmod.fft.ifft( npmod.fft.ifftshift(
+            #                npmod.fft.fftshift( npmod.fft.fft(
+            #                    array_in , axis=0 ), axes=0) * H,
+            #                        axes=0 ), axis=0 )
+
+            array_out = signal.sosfilt(sos, array_in, axis=0)
 
             # store in the output
             array_name_out = array_name_in + '_' + self.filter_type

```

taobrienlbl:
> @taobrienlbl I tried using scipy signal butter and sos filtering as this was an easy change to make. there is really bad distortion at the start of the filtered signal. 
I think that this is expected for this type of filter.  AFAIK, this type of filter moves forward through a dataset (analogous to how a filter from an incoming microphone signal might filter the incoming signal on the fly).  I believe it comes about because it is assumed the signal is 0 prior to the start.

For the types of applications we'd be working on, I think this artifact would be a very minor component of the signal.  I would personally discard the first bit of my data if I were using this function.

Btw, I think this makes an excellent example even if it can't immediately use the GPUs, since it demonstrates wrapping existing python libraries to do timeseries analysis with TECA.

taobrienlbl:
> Notice the errors on the left side of the filtered signal. The MSE for the sos based method is 0.52 and for my fft based method 0.0002
I'll also add that it looks like the distortion at the beginning has a duration of about 2*`1/cutoff_frequency`, which is approximately the width of the filter in temporal space.  So as long as the length of the dataset is >> than `1/cutoff_frequency`, the problem is of minimal importance, unless there's something interesting to analyze right at the beginning of the dataset.

burlen:
prw_high_pass 0132

burlen:
the high pass filter seems to capture the ar features! https://drive.google.com/file/d/1zAF4iANF4MwRTsllnYYqPa7gZewuG95h/view?usp=sharing

burlen:
bw_filter_response
Because the sampling rate is fixed and out of our control, and the time scales are long, when using this filter, one should make plot such as this to verify the filter response is reasonable. For instance in the above case the 2nd order low pass never hits zero over the filter window

burlen:
Using collective buffering and appropriate Lustre striping has a big impact with spatial parallelism
spectral_filter_65y_ecmwf_runtime_16n
spectral_filter_65y_ecmwf_bandwidth_16n

PR #723, opened 2022-09-01
Temporal reduction

amandasd:


alg/teca_temporal_reduction.cxx
```diff
@@ -0,0 +1,2443 @@
...
+      = std::dynamic_pointer_cast<const teca_cartesian_mesh>(data_in[n_data-1]);
+    p_teca_cartesian_mesh mesh_out = teca_cartesian_mesh::New();
+    mesh_out->shallow_copy(
+      std::const_pointer_cast<teca_cartesian_mesh>(mesh_in));
+    p_teca_array_collection arrays_out = mesh_out->get_point_arrays();
+
+    size_t n_array = this->point_arrays.size();
+
+    // accumulate incoming values
+    for (int i = n_data-2; i >= 0; --i)
```
burlen, line 2368: What happens if the number of input datasets is < 2? this is the source of the crash on `PreIndust.ar_tag.Guan_Waliser_v2` which has only 1 time step in the last interval.

We have 3 options.
1. error out when the number of inputs is less than 2.
2. handle the case where the number of inputs < 2 by passing the input through, with type conversion applied when needed. In this case set the array to all zero or the _FillValue if one is provided.
3. implement a "complete interval only" mode in the iterators, activate it for the temporal reduction app, and error out when the number of inputs is < 2.

Option 1 could be frustrating for users, you could end up processing 30 years of data then error out in the very last time step. This is an edge case so I doubt it will come up in practice very much.  Option 2 handles the issue but puts owness on the user to interpret the result. Option 3  has the advantage in that one's always sure that all the data has been weighted the same.

I'm leaning toward 2. @taobrienlbl  thoughts on this?

taobrienlbl: I'd lean toward 2 as well.  That's the behavior I would intuitively expect.

amandasd: Good catch @burlen! I will work on this. Thanks!

alg/teca_temporal_reduction.cxx
```diff
@@ -2355,21 +2369,78 @@ const_p_teca_dataset teca_temporal_reduction::execute(
...
+                    }
+#endif
+                    NT *p_red_array = sp_red_array.get();
+
+                    for (unsigned int i = 0; i < n_elem; ++i)
+                        p_red_array[i] = (this->fill_value != -1) ? this->fill_value : 0.;
+                )
+                arrays_in->set(array, red_array);
+            }
+        }
```
burlen, line 2443: handle the n_data < 2 case outside the for loop.

minor: n_data has 64 bits, better to use a 64 bit integer in the loop counter.

amandasd: Do you mean `unsigned long` instead of `size_t`?

burlen: nope, I mean use long (64 bit) instead of int (32 bit) in the loop count because n_data is size_t (64 bit). when mixing 32 and 64 bit types there can be an overflow on the 32 bit type.

burlen: also, I suspect there will be a compiler warning generated for this code because an overflow can also happen when mixing signed and unsigned types. Because of the intial loop count (i = n_data - 2) could be negative in some case it's probably better to use a signed data type for n_data such as ssize_t or long as well.

alg/teca_temporal_reduction.cxx
```diff
@@ -0,0 +1,2508 @@
...
+                if (device_id >= 0)
+                {
+                    auto sp_red_array = static_cast<TT*>(red_array.get())->get_cuda_accessible();
+                }
+                else
+                {
+#endif
+                    auto sp_red_array = static_cast<TT*>(red_array.get())->get_cpu_accessible();
+#if defined(TECA_HAS_CUDA)
+                }
```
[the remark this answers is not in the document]
amandasd: Thanks! I will fix it now.

burlen:
@amandasd this is not compiling with CUDA enabled. Would you please test the changes on Perlmutter? You can use the CFS file system  while scratch is down. The procedure will be exactly the same as using scratch. Also would please see if you can clean up history a bit? Commit messages should briefly describe the changes and which classes were modified. Minor changes and series of changes to a new class could be squashed. It's ok not to squash if there's some reason. Let me know if you want to sync up on this.
```
/home/bloring/work/teca/teca/alg/teca_temporal_reduction.cxx(2409): error: identifier "sp_red_array" is undefined
```

amandasd:
> 

Sorry about that. I just fixed it.

burlen:
this is not really about squashing in and of itself. This is about grouping things logically (independent changes get grouped) and documenting so that other developers can quickly get the gist of the changes in a given commit when debugging. Some specifics follow. While not exhaustive, I'm hoping that what follows will help clarify 

I'd like to see commit messages that 

1. Tell what classes/apps etc were modified 
2. have a brief description of the modifications

Here's a specific example from the log that could be improved:
```
commit c7a202c132f4d43f4b101cd1d91ca296eecf78f0
Author: amanda <amanda.dufek@gmail.com>
Date:   Fri Jul 15 16:01:17 2022 -0700

    Added spatial parallelism
    
    Added spatial parallelism
    
    Updated spatial parallelism
    
    Resolving conflicts
    
    Minor changes on documentation
    
    Documentation text
```
This doesn't say which class(es) were modified. Once merged and later on say in 6 mos, it will not be obvious. Notes like "resolving conflicts" and "minor changes" don't provide useful info. Multiple repeated lines don't provide useful info. 

Use the following template for commits
1. a short <= 80 ish char title that names the class and captures the gist of the changes. some times this is enough. if not see step 2
2. if the changes were significant such that the title doesn't provide enough detail : skip a line and go into detail. this can be a sentence or two up to as much info as you think necessary. I often break it into a bullet list.

the point of this info is to help us later on, when we are debugging, have a rough idea what was done in this commit so we can quickly rule it out as the source of the bug we are investigating or quickly decide if we should look more closely. The idea is that someone other than yourself would have a reasonable chance of using the info.

When working on a large change (as you've done here) it is common to make a ton of small commits, and that's fine. When preparing to merge how do you decide weather or not to squash a series of commits? As a rule of thumb it's best to squash the intermediate work where the code was not fully functional or not working correctly into the commit that added the class. Go into the process with that goal in mind.

There are a number of good reasons not to squash commits, for instance in might be useful to have a new commit for a major change that follows a commit where things are working. In that case squash down to just commits for major changes that you want to go back to in the future for purposes of having a reference when debugging. This is just one example. An other example is squashing through a move can be problematic and not worth the effort. The point is to squash things down that you don't have a good reason to keep.

Here's a specific example:
```
commit bdb977dd4a2c2cd271456f7e80166ddde6af8189
Author: amanda <amanda.dufek@gmail.com>
Date:   Mon Jun 27 15:55:24 2022 -0700

    Changes in the n_steps iterator
```
Since the C++ n_steps iterator is a new class you add in this branch, it would likely make sense to squash this, and other intermediate changes to that class into the commit that adds the class. Note: this is also an example of insufficient detail in the message.

In general you want to use commits to group related work. The reasons for this include making it easy to revert  a buggy commit without losing unrelated work.

Here's an example:
```
$git show 3cccc1 --name-only
commit 3cccc108e689ed571a624a5d05fa68d60c72b681
Author: amanda <amanda.dufek@gmail.com>
Date:   Thu Jul 21 11:06:33 2022 -0700

    Added a test for the n_steps iterator
    
    Resolving warnings
    
    Resolving warnings
    
    Resolving warnings

alg/teca_temporal_reduction.cxx
data/teca_calendar_util.cxx
data/teca_calendar_util.h
test/apps/CMakeLists.txt
test/apps/test_temporal_reduction_app.sh
test/apps/test_temporal_reduction_app_mcf.sh
```
You added a test and cleaned up warnings. If you accidentally introduced a bug when cleaning up the warnings we would not want to revert the test. In this case it would be best to split this commit into 2. One adds the test the other cleans up the warnings.

Another reason for grouping logically is to document important changes in an obvious way. Here's an example that could go either way
```
$ git show 310546a --name-only
commit 310546aa4eac5f4fad56037a27bc2879d6b838b3
Author: amanda <amanda.dufek@gmail.com>
Date:   Fri Aug 19 09:15:08 2022 -0700

    Renamed the original python implementation
    
    Increased the size of the class_name variable from 64 to 96
    
    Updated class name in test/python/test_temporal_reduction.py
    
    Updated class name in test/python/test_custom_temporal_reduction.py

alg/teca_python_temporal_reduction.py
apps/teca_python_temporal_reduction.in
core/teca_threaded_programmable_algorithm.h
test/apps/CMakeLists.txt
test/apps/test_python_temporal_reduction_app.sh
test/apps/test_python_temporal_reduction_app_mcf.sh
test/python/test_custom_temporal_reduction.py
test/python/test_temporal_reduction.py
```
This commit in addition to renaming the python implementation makes an important change to one of teca's core classes: teca_threaded_programmable_algorithm. In this case you increased the size of the class name variable. While it is OK to include this change as part of this larger commit, it might be worth having it in it's own commit, so that it's obvious and that we don't accidentally revert it.

I hope this helps clarify. Resolving these issues could probably be done in an hour or so, if you;d like some help with git side of it, let's meet in person at building 59.

amandasd (commented):
Hi @burlen. I think we should include the changes made to test_python_temporal_reduction_app.sh also in test_python_temporal_reduction_app_mcf.sh, shouldn't we?

burlen:
@amandasd yes! good catch. Thank you

burlen (approved):
@amandasd thank you for all the work on this!

burlen:
@amand please go ahead and merge! Once it's in I'll  start deployments on cori perlmutter

PR #742, opened 2023-01-27
regional moisture flux

burlen:
The attached pdf documents the approach
[conus_moisture_flux (7).pdf](https://github.com/LBL-EESA/TECA/files/10523872/conus_moisture_flux.7.pdf)

### DIFF
--- a/alg/teca_regional_moisture_flux.cxx
+++ b/alg/teca_regional_moisture_flux.cxx
@@ -1,0 +1,708 @@
+#include "teca_regional_moisture_flux.h"
+
+#include "teca_cartesian_mesh.h"
+#include "teca_table.h"
+#include "teca_array_collection.h"
+#include "teca_variant_array.h"
+#include "teca_variant_array_impl.h"
+#include "teca_variant_array_util.h"
+#include "teca_metadata.h"
+#include "teca_array_attributes.h"
+
+#include <algorithm>
+#include <iostream>
+#include <string>
+#include <set>
+#include <cmath>
+
+#if defined(TECA_HAS_BOOST)
+#include <boost/program_options.hpp>
+#endif
+
+#if defined(TECA_HAS_CUDA)
+#include "teca_cuda_util.h"
+#include <cuda.h>
+#include <cuda_runtime.h>
+#include <thrust/device_vector.h>
+#include <thrust/functional.h>
+#include <thrust/transform_reduce.h>
+#include <thrust/iterator/zip_iterator.h>
+
+#endif
+
+using namespace teca_variant_array_util;
+using allocator = teca_variant_array::allocator;
+
+//#define TECA_DEBUG
+
+namespace
+{
+// CPU codes
+//
+
+namespace cpu
+{
+// **************************************************************************
+template <typename coord_t>
+void transform_coordinates(coord_t *theta, coord_t *phi, coord_t *sin_phi,
+    const coord_t *lon, const coord_t *lat, unsigned long nlon, unsigned long nlat)
+{
+    // transform the cooridnates
+    coord_t rad_deg = M_PI / 180.0;
+
+    for (unsigned long i = 0; i < nlon; ++i)
+    {
+        theta[i] = lon[i] * rad_deg;
+    }
+
+    for (unsigned long j = 0; j < nlat; ++j)
+    {
+        coord_t phi_j = (90.0 - lat[j]) * rad_deg;
+
+        phi[j] = phi_j;
+
+        sin_phi[j] = sin(phi_j);
+    }
+}
+
+
+// **************************************************************************
+template <typename coord_t, typename data_t>
+void divergence(data_t *div, const data_t *ivt_u, const data_t *ivt_v,
+    const coord_t *theta, const coord_t *phi, const coord_t *sin_phi,
+    unsigned long nlon, unsigned long nlat)
+{
+    unsigned long nlatm1 = nlat - 1;
+    unsigned long nlonm1 = nlon - 1;
+
+    // compute the divergence at every point
+    for (unsigned long j = 1; j < nlatm1; ++j)
+    {
+        for (unsigned long i = 1; i < nlonm1; ++i)
+        {
+            unsigned long q = j*nlon + i;
+
+            // div = d / d theta ( ivt_theta )
+            //     + d / d phi ( ivt_phi sin(phi) )
+
+            // forward difference. using a centered difference here resulted in
+            // a larger overall error.
+            data_t dthe = theta[i + 1] - theta[i];
+            data_t dphi = phi[j + 1] - phi[j];
+
+            data_t divt_the = ivt_u[q + 1] - ivt_u[q];
+            data_t divtsin_phi =  ivt_v[q + nlon] * sin_phi[j + 1] - ivt_v[q] * sin_phi[j];
+
+            div[q] = divt_the / dthe + divtsin_phi / dphi;
+        }
+    }
+}
+
+// **************************************************************************
+template <typename coord_t, typename data_t>
+void integrate(data_t *cell_int, const data_t *div,
+    const coord_t *theta, const coord_t *phi,
+    unsigned long nlon, unsigned long nlat)
+{
+    unsigned long nlatm1 = nlat - 1;
+    unsigned long nlonm1 = nlon - 1;
+
+    // integrate cell by cell using the trapezoid rule
+    for (unsigned long j = 1; j < nlatm1; ++j)
+    {
+        for (unsigned long i = 1; i < nlonm1; ++i)
+        {
+            unsigned long q = j*nlon + i;
+
+            // area of the cell
+            data_t dthe = theta[i + 1] - theta[i];
+            data_t dphi = phi[j + 1] - phi[j];
+            data_t dA = dthe * dphi;
+
+            // height of the surface above the cell. using trapezoid here
+            // resulted in a larger overall error.
+            cell_int[q] = div[q] * dA;
+        }
+    }
+}
+
+// ***************************************************************************
+template <typename data_t, typename mask_t>
+void sum_if(double &moisture_flux, const data_t *cell_int,
+    const mask_t *reg_mask, unsigned long nelem)
+{
+    moisture_flux = 0.0;
+    for (unsigned long q = 0; q < nelem; ++q)
+    {
+        moisture_flux += reg_mask[q] ? cell_int[q] : data_t();
+    }
+}
+
+// ***************************************************************************
+int dispatch(double &moisture_flux,
+    const const_p_teca_variant_array &lon, const const_p_teca_variant_array &lat,
+    const const_p_teca_variant_array &ivt_u, const const_p_teca_variant_array &ivt_v,
+    const const_p_teca_variant_array &region_mask,
+    unsigned long nlon, unsigned long nlat)
+{
+    NESTED_VARIANT_ARRAY_DISPATCH_FP(lon.get(), _COORDS,
+        assert_type<TT_COORDS>(lat);
+
+        unsigned long nelem = nlat*nlon;
+
+        auto [splon, plon] = get_host_accessible<CTT_COORDS>(lon);
+        auto [splat, plat] = get_host_accessible<CTT_COORDS>(lat);
+
+        auto [theta, ptheta] = New<TT_COORDS>(nlon);
+        auto [phi, pphi] = New<CTT_COORDS>(nlat);
+        auto [sin_phi, psin_phi] = New<TT_COORDS>(nlat);
+
+        transform_coordinates(ptheta, pphi, psin_phi, plon, plat, nlon, nlat);
+
+        NESTED_VARIANT_ARRAY_DISPATCH_FP(ivt_u.get(), _DATA,
+            assert_type<TT_DATA>(ivt_v);
+
+            auto [spivt_u, pivt_u] = get_host_accessible<CTT_DATA>(ivt_u);
+            auto [spivt_v, pivt_v] = get_host_accessible<CTT_DATA>(ivt_v);
+
+            auto [div, pdiv] = New<TT_DATA>(nelem, NT_DATA());
+            auto [cint, pcint] = New<TT_DATA>(nelem, NT_DATA());
+
+            // divergance
+            divergence(pdiv, pivt_u, pivt_v, ptheta, pphi, psin_phi, nlon, nlat);
+
+            // integrate cell by cell
+            integrate(pcint, pdiv, ptheta, pphi, nlon, nlat);
+
+            NESTED_VARIANT_ARRAY_DISPATCH(region_mask.get(), _MASK,
+
+                auto [spreg_mask, preg_mask] = get_host_accessible<CTT_MASK>(region_mask);
+
+                // reduce/sum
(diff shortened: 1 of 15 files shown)


burlen:
using streams can be a way to speed up the execution of many small kernel invocations
shape_file_mask_streams

alg/teca_regional_moisture_flux.cxx
```diff
@@ -0,0 +1,690 @@
...
+    //     + d / d phi ( ivt_phi sin(phi) )
+    //
+    // dA = d theta d phi
+    //
+
+    // area of the cell
+    data_t dthe = theta[i + 1] - theta[i];
+    data_t dphi = phi[j + 1] - phi[j];
+
+    data_t dA = dthe * dphi;
```
taobrienlbl, line 293: shouldn't there be a multiplication by sin(phi[j]) here to account for convergence of the meridians (i.e., dA should go to 0 near the poles)?

alg/teca_regional_moisture_flux.cxx
```diff
@@ -0,0 +1,690 @@
...
+            unsigned long q = j*nlon + i;
+
+            // area of the cell
+            data_t dthe = theta[i + 1] - theta[i];
+            data_t dphi = phi[j + 1] - phi[j];
+            data_t dA = dthe * dphi;
+
+            // height of the surface above the cell. using trapezoid here
+            // resulted in a larger overall error.
+            cell_int[q] = div[q] * dA;
```
taobrienlbl, line 124: shouldn't there be a multiplication by sin(phi[j]) here to account for convergence of the meridians (i.e., dA should go to 0 near the poles)?

alg/teca_regional_moisture_flux.cxx
```diff
@@ -0,0 +1,690 @@
...
+
+            // forward difference. using a centered difference here resulted in
+            // a larger overall error.
+            data_t dthe = theta[i + 1] - theta[i];
+            data_t dphi = phi[j + 1] - phi[j];
+
+            data_t divt_the = ivt_u[q + 1] - ivt_u[q];
+            data_t divtsin_phi =  ivt_v[q + nlon] * sin_phi[j + 1] - ivt_v[q] * sin_phi[j];
+
+            div[q] = divt_the / dthe + divtsin_phi / dphi;
```
taobrienlbl, line 96: should there be a (1/(r_earth*sin_phi[j])) multiplication here?  If it's somehow taken care of elsewhere in the code, it might be worth adding comments here about why/how

alg/teca_regional_moisture_flux.cxx
```diff
@@ -0,0 +1,690 @@
...
+
+    // forward difference. using a centered difference here resulted in
+    // a larger overall error.
+    data_t dthe = theta[i + 1] - theta[i];
+    data_t dphi = phi[j + 1] - phi[j];
+
+    data_t divt_the = ivt_u[q + 1] - ivt_u[q];
+    data_t divtsin_phi =  ivt_v[q + nlon] * sin_phi[j + 1] - ivt_v[q] * sin_phi[j];
+
+    div[q] = divt_the / dthe + divtsin_phi / dphi;
```
taobrienlbl, line 259: should there be a (1/(r_earth*sin_phi[j])) multiplication here?  If it's somehow taken care of elsewhere in the code, it might be worth adding comments here about why/how

alg/teca_regional_moisture_flux.cxx
```diff
@@ -0,0 +1,690 @@
...
+
+            NESTED_VARIANT_ARRAY_DISPATCH(region_mask.get(), _MASK,
+
+                auto [spreg_mask, preg_mask] = get_cuda_accessible<CTT_MASK>(region_mask);
+
+                // sum over the masked region
+                moisture_flux = sum_if(pcint, preg_mask, data_nelem);
+
+                NT_DATA Re = 6378100.;
+                moisture_flux *= -Re;
```
taobrienlbl, line 425: I made two comments above about a multiplication by (1/(Re*sin_phi[j]).  I see here that it looks like this is being taken care of here, which might be worth commenting on at the above locations.  I assume its to avoid multiplying everything by Re in the inner loop and instead doing it outside the loop?  But there's still the question of the 1/sin_phi -- maybe you did some transformation to avoid that term?

alg/teca_regional_moisture_flux.cxx
```diff
@@ -0,0 +1,690 @@
...
+
+    std::set<std::string> arrays;
+    if (req.has("arrays"))
+        req.get("arrays", arrays);
+
+    arrays.insert(this->ivt_u_variable);
+    arrays.insert(this->ivt_v_variable);
+    arrays.insert(this->region_mask_variable);
+
+    // TODO if the valid value masks for IVT are available request them too
```
taobrienlbl, line 588: flagging this TODO in a comment so it doesn't get missed

alg/teca_regional_moisture_flux.cxx
burlen: > should there be a (1/(r_earth*sin_phi[j])) multiplication here? If it's somehow taken care of elsewhere in the code, it might be worth adding comments here about why/how

There's also a factor of r_earth^2*sin_phi from dA that cancels with the factor you mention. After a factor of r_earth remains but I pulled this out of the integral and applied at the very end. For reference see equation 8 in the overleaf. Good idea to improve the documentation, I will add equation 8 to the doxygen.

alg/teca_regional_moisture_flux.cxx
burlen: > shouldn't there be a multiplication by sin(phi[j]) here to account for convergence of the meridians (i.e., dA should go to 0 near the poles)?

It cancels with the factor of 1/(r_earth*sin(phi)) from the divergence. See eqn 8 in the overleaf.

alg/teca_regional_moisture_flux.cxx
burlen: > should there be a (1/(r_earth*sin_phi[j])) multiplication here? If it's somehow taken care of elsewhere in the code, it might be worth adding comments here about why/how

explained above.

alg/teca_regional_moisture_flux.cxx
burlen: > shouldn't there be a multiplication by sin(phi[j]) here to account for convergence of the meridians (i.e., dA should go to 0 near the poles)?

explained above

alg/teca_regional_moisture_flux.cxx
burlen: > I made two comments above about a multiplication by (1/(Re*sin_phi[j]). I see here that it looks like this is being taken care of here, which might be worth commenting on at the above locations. I assume its to avoid multiplying everything by Re in the inner loop and instead doing it outside the loop? But there's still the question of the 1/sin_phi -- maybe you did some transformation to avoid that term?

Yes, `/(r_earth*sin(phi)` from the divergence cancels with the `r_earth^2*sin(phi)` from the dA. After a factor of `r_earth` remains. I pulled it out of the integral and apply it here, the very last step of  the calc. This all comes from eqn 8 in the overleaf, I will add this to the doxygen in the header file.

I can see how this is confusing the way the calculation is split up here into multiple steps. I split things up this way due to load balancing concerns on the GPU.

alg/teca_regional_moisture_flux.cxx
burlen: @taobrienlbl Question, do we need to handle missing values in IVT field? I'm guessing it would be a good idea, but maybe save it for future work?

alg/teca_regional_moisture_flux.cxx
taobrienlbl: got it, thanks.  I was referring to another source for the spherical coordinate divergence, just for independence, but yes, I agree with this cancellation.

alg/teca_regional_moisture_flux.cxx
taobrienlbl: makes sense, thanks!

alg/teca_surface_integral.cxx
```diff
@@ -0,0 +1,712 @@
...
+
+                NESTED_VARIANT_ARRAY_DISPATCH(region_mask.get(), _MASK,
+
+                    auto [spreg_mask, preg_mask] = get_cpu_accessible<CTT_MASK>(region_mask);
+
+                    // reduce/sum
+                    double reg_total;
+                    sum_if(reg_total, pcint, preg_mask, nelem);
+
+                    // - because limits of integration are flipped in our coordinate system
```
taobrienlbl, line 150: might be nice to add this comment in the ivt flux calculation too

alg/teca_surface_integral.h
```diff
@@ -0,0 +1,117 @@
...
+ * \f}
+ *
+ * Note: the minus sign out front is because we flip the limits of integration
+ * when going from latitude to \f$phi\f$ coordinates.
+ *
+ * Requirements:
+ *
+ * 1. A list of surface variables
+ * 2. A list of names to use when packaging the results
+ * 3. A region mask a 2D where non-zero entries defining
```
taobrienlbl, line 40: Consider rewriting as "A region mask: a 2D array where ..."

alg/teca_table_reduce.h
```diff
@@ -13,9 +13,9 @@
 TECA_SHARED_OBJECT_FORWARD_DECL(teca_table_reduce)
 
 /// A reduction on tabular data over time steps.
-/**
- * Tabular data from each time step is collected and
- * concatenated into a big table.
+/** Tabular data from is concatenated into a single table. Threading results in
```
taobrienlbl, line 16: "Tabular data from is" --> "Tabular data from each timestep is" ?

apps/teca_regional_moisture_flux.cpp
```diff
@@ -0,0 +1,612 @@
...
+    // shape file mask
+    std::string region_name = opt_vals["region_name"].as<string>();
+
+    mask->set_input_connection(mask_mesh->get_output_port());
+    mask->set_shape_file(opt_vals["shape_file"].as<string>());
+    mask->set_mask_variable(region_name + "_mask");
+
+    mask_cache->set_input_connection(mask->get_output_port());
+
+    mask_regrid->set_input_connection(0, head->get_output_port());
```
taobrienlbl, line 444: At some point (in another PR), it might be worth separating out the functionality being used here in  `teca_cartesian_mask_regrid`: maybe a new algorithm like `teca_pipeline_join`.  The only reason would be to make it more clear to a power user how to combine the dataset cache.  On a technical side, we also might find some use cases where we want to have multiple branches of a pipeline upstream of an algorithm (maybe a cf_reader, a table_reader, and a dataset cache).  The `teca_pipeline_join` algorithm could be configured to accept a flexible number of input connections.

alg/teca_regional_moisture_flux.cxx
burlen: Hmm. Might make sense to rename this algorithm to `teca_atmospheric_flux` and use some generic names like vector_u and vector_v instead of ivt_u and v?

burlen: Oh wait, I think we've made some simplifications in our derivation that only apply to IVT. So I guess that's not going to work.

apps/teca_regional_moisture_flux.cpp
burlen: Yes, assuming the meshes have the same spatial coordinate system, then we could then join any number of pipeline branches. Great idea. 

Also, we really need more documentation of what's already implemented and how to use.

core/teca_index_reduce.cxx
```diff
@@ -143,6 +146,14 @@ void teca_index_reduce::get_properties_description(const std::string &prefix,
         + (prefix.empty()?"teca_index_reduce":prefix));
 
     opts.add_options()
+        TECA_POPTS_MULTI_GET(extent_type, prefix, extent,
+            "an index space spatial subset of the data to request"
+            " [i0, i1, j0, j1, k0, k1] (optional)")
+        TECA_POPTS_MULTI_GET(bounds_type, prefix, bounds,
+            "a world space subset of the data to request"
+            " [x0, x1, y0, y1, z0, z1] (optional)")
```
taobrienlbl, line 154: This is a nice new feature!

io/teca_shape_file_mask.h
```diff
@@ -12,7 +12,13 @@
 TECA_SHARED_OBJECT_FORWARD_DECL(teca_shape_file_mask)
 
 /// Generates a valid value mask defined by regions in the given ESRI shape file
-/** This algorithm is a source and has no inputs. A
+/** The teca_shape_file_mask generates a mask on a 2D mesh where mesh points
+ * are assigned a value of 1 where they fall inside any one of the polygons
+ * defined in the named ESRI shape file, and 0 everywhere else. The input mesh
```
taobrienlbl, line 17: Based on the code here, it looks like it is assumed that the polygons in the shape file are in [the EPSG:4326 projection](https://spatialreference.org/ref/epsg/4326/) (or WSG84).  This should be reported by the CRS variable in the shapefile, so it should be possible to check.  Maybe this is checked somewhere that I'm not immediately seeing though?

If I'm right about this, I suggest changing the documentation to indicate this (there are many shapefiles out in the wild that are in other projections, which would not work here) and possibly also implementing a check on the CRS value.  In a future pair of PRs, it might be worth considering (1) adding [GDAL](https://gdal.org/) to the superbuild, and (2) [using GDAL to transform](https://gdal.org/api/ogrspatialref.html) between the shapefiles coordinate system and EPSG:4326 coordinate system.

taobrienlbl (approved):
I've finally finished the review.  A lot of work went into this PR!  The code all looks good to me, and the tests are great.  Thanks also for clarifying the comments about the Re and sin terms.  This is solid code.  None of the comments I made should prevent this being merged; they're just thoughts for future work.  Thanks @burlen!

io/teca_shape_file_mask.h
burlen: You're right, except that we can handle both -180 to 180 and 0 to 360 in lon.  I've encountered a number of shape files in other coordinate systems and with our current implementation there's no indication of a problem except a mask of all zeros. Then I've used other tools to dump the shapefile inorder to investigate further.

The library we're currently using http://shapelib.maptools.org/ may not support the CRS field. In fact I can find no mention of the term "coordinate system" in the shapefile format specification http://dl.maptools.org/dl/shapelib/shapefile.pdf Where is the CRS field documented?

Thanks for the pointer to gdal, added an issue for this #744

apps/teca_regional_moisture_flux.cpp
burlen: Captured in issue #743

alg/teca_surface_integral.h
burlen: done

alg/teca_table_reduce.h
burlen: improved this.

burlen:
@taobrienlbl I've merged all the supporting work and all that remains here are the commits adding moisture flux, surface integral stages and apps.

Could we potentially come up with a plan to finish up the validation of these? For instance what additional validations could be done? So far tests have been limited to ECMWF IFS, is there another dataset that we could use to verify that it's not something funky with this particular dataset?

PR #774, opened 2023-08-22
profile cpp reduction wo. I/O

burlen:


burlen:
gpu_threading
cpu_threading

burlen:
@amandasd I would like to merge these changes into your branch. let me know if that's OK

amandasd:
I just wondering if we really need to duplicate the average_finalize kernel.
Where in the code did you see that count is stored on the host?

burlen:
> @burlen I just wondering if we really need to duplicate the average_finalize kernel. Where in the code did you see that count is stored on the host?

Ah, it looks like you may have fixed this issue as well. The problem (still exists on the develop branch) was that when missing values were not used the count was allocated on the host, then used in the finalize kernel. This caused a segv. It was not encountered before since the apps always have missing values. My new test does not use them so it crashed.

https://github.com/LBL-EESA/TECA/blob/c78d39b05623fb33cc5487e038a6441ce8a44c55/alg/teca_temporal_reduction.cxx#L886

It's great that you fixed this in your recent work, I will remove that commit

burlen:
@amandasd I removed the commit modifying the count and finalize kernels. That was not needed anymore since you fixed it.

burlen:
@amandasd I merged this onto your branch. I'd recommend getting these chanegs asap, since it may solve the threading issue you saw and will be easier to resolve conflicts  now, and be harder if you wait. I'm happy to help resolve conflicts or git issues if there are any

PR #777, opened 2023-09-09
Cfr page locked memory

burlen:
improves overall performance of the temporal reduction application by replacing cudaMallocHost w/ malloc allocators in the cf_reader (I made the change in hamr as well for device to host transfers)


burlen:
@amandasd you'll want these before doing more profiling. it should not conflict with your work since it's in the reader. also made the change in HAMR so update submodule after

PR #778, opened 2023-09-11
temporal_reduction construct output at end and const correctness

burlen:
* don't modify the input data
* don't construct the output in every pass. instead defer until the calculation is complete
* apply check for floating point type during initialization to avoid a deep copy of every input


burlen:
@amandasd a few improvements. what do you think?

PR #793, opened 2024-01-25
Force BARD default for periodic_in_x

### LINKED ISSUE #792: TECA BARD does not default to periodic_in_x=1 even though the documentation indicates it should
The help for `teca_bayesian_ar_detect --help` indicates the following:

```
  --periodic_in_x arg (=1)
    
    Flags whether the x dimension (typically longitude) is periodic.
```

This suggests that by default, TECA BARD should assume that datasets are periodic, which is consistent with the underlying assumption in BARD that the data are global.  The actual default value is zero because [`teca_bayesian_ar_detect.cxx`](https://github.com/LBL-EESA/TECA/blob/b77e030197f9ca417cc920aa91ee96e921cbeb41/apps/teca_bayesian_ar_detect.cpp#L343) defers setting the mesh periodicity if the default value is provided on the command line; this delegates selection of the default value to the CF reader, which has a default of 0.

Datasets produced with TECA BARD that did not explicitly set the `--periodic_in_x` flag will therefore not treat ARs crossing the periodic boundary correctly.

taobrienlbl:
Fixes #792

### DIFF
--- a/apps/teca_bayesian_ar_detect.cpp
+++ b/apps/teca_bayesian_ar_detect.cpp
@@ -184,6 +184,7 @@ int main(int argc, char **argv)
 
     p_teca_multi_cf_reader mcf_reader = teca_multi_cf_reader::New();
     mcf_reader->get_properties_description("mcf_reader", advanced_opt_defs);
+    mcf_reader->set_periodic_in_x(1);
 
     p_teca_valid_value_mask vv_mask = teca_valid_value_mask::New();
     vv_mask->get_properties_description("vv_mask", advanced_opt_defs);


burlen:
Don't do it this way. It will break how advanced options work. basic options are short hand for the most commonly used properties while advanced options expose every modifiable property of every pipeline stage in the app, in this case the basic option is `--periodic_in_x` and advanced option is `--mcf_reader::periodic_in_x`. The idea is that an app force override defaults when it constructs the class and sets the basic option default the same. Then the app parses/applies the advanced options . Then the basic flags are parsed and applied, but only if they are not defaulted. This sequence gives the basic options precedence (because they are shorter to type/easier to use) but still lets one specify either advanced form or basic form and get the same result.

If the basic options were always applied, as in this change, they wipe out what ever was passed as an advanced option. 

instead you could add `mcf_reader->set_periodic_in_x(1);` at line 187.

taobrienlbl:
thanks @burlen - that worked nicely w/o overriding the advanced option system

burlen:
> thanks @burlen - that worked nicely w/o overriding the advanced option system

For context the reason that it was not set before in the application on the MCF reader is that doing so will take precedence over the MCF file. Then you can't set it in the MCF file (if you do it is silently ignored). The down side of the original approach is that then people have to remember to set it in the file if they want it. Also, note that setting on the MCF reader algorithm properties are applied to all sub-readers. This is what you want in this case, but for other properties it is not.

taobrienlbl:
Thanks @burlen - that's helpful context.  So it wasn't so much a bug as a misunderstood feature.  I think the current change is the sensible way to proceed, but we can add a warning if the option is picked up in the MCF file.

burlen:
I agree, this is an improvement.

PR #798, opened 2024-06-17
Draft: Add support for reading variables in groups.

ghweber:
Add support to read variables that are not in the `root` group but instead a group.

Refer to variables by `<groupname>/<varname>' in `.mfc` file.

Limitations:
* Not sure if it is ok to cache `cf_parent_id` or if a group `ncid` can change when a file is closed and re-opened.
* Currently only supports variables in top-level groups. Need to add recursion to support groups at arbitrary depth in hierarchy. 
*  No support for groups in `teca_wrf_reader`, yet.

### DIFF
--- a/io/teca_array_collection_reader.cxx
+++ b/io/teca_array_collection_reader.cxx
@@ -313,7 +313,7 @@ teca_metadata teca_array_collection_reader::get_output_metadata(unsigned int por
         {
             std::string name;
             teca_metadata atts;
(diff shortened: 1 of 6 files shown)


taobrienlbl:
thanks for the work so far!  FYI, it's very unlikely that WRF files would have HDF5 groups, so I think it is okay to leave `teca_wrf_reader` alone for now

ghweber:
Replaced by #799. (Needed to move to branch in main repository instead of fork)

PR #799, opened 2024-06-18
Add support for reading variables in groups.

ghweber:
Add support to read variables that are not in the `root` group but instead a group. (Replaces #798)

Refer to variables by `<groupname>/<varname>' in `.mfc` file.

Limitations:
* Not sure if it is ok to cache `cf_parent_id` or if a group `ncid` can change when a file is closed and re-opened.
* Currently only supports variables in top-level groups. Need to add recursion to support groups at arbitrary depth in hierarchy. 
*  No support for groups in `teca_wrf_reader`, yet. (may not be needed according to Travis)


taobrienlbl (approved):
@ghweber - the changes look great, all tests pass in TravisCI, and the new tests pass on my home machine.  This looks good to merge!

PR #802, opened 2024-10-28
Ensemble support in reader and support for groups in Netcdfwriter

ghweber:
Superseded by #803


io/teca_cf_reader.cxx
```diff
@@ -1721,6 +1734,16 @@ const_p_teca_dataset teca_cf_reader::execute(unsigned int port,
                 // select the requested time step
                 // subset point centered variables based on the incoming requested
                 // extent.
+
+                 // TODO/FIXME: Currently we assume that the ensemble is the first
```
taobrienlbl: before we close this PR, let's make sure to open an issue to generalize this (this will end up being a useful feature, btw)

io/teca_cf_layout_manager.cxx
```diff
@@ -15,6 +15,79 @@
 #include <cerrno>
 #include <string>
 
+namespace
+{
+    int full_path_varname_to_group_id_and_name_in_group(int file_id,
+        const std::string& full_path_varname, int &parent_id,
+        std::string& name_in_group)
+    {
+        std::cout << "full_path_varname_to_group_id_and_name_in_group: " << file_id << " " << full_path_varname;
```
taobrienlbl: I assume the `std::cout` lines in this subroutine are debugging artefacts?  These should probably be removed or changed to only write if the `verbose` flag is set for the writer

ghweber: I already removed this but it seems I forgot to commit that change. Planning on just removing them.

ghweber: @taobrienlbl , I had removed that and performed a bit more clean-up but forgotten to commit. Can you please take another close look at the change, please?

io/teca_cf_reader.cxx
ghweber: If I read it correctly, TECA already assumes an implicit order of dimensions. For example it seems to assumes it is always `time`, `z`, `y`, `x` if I read everything correctly. Checking for dimension names is in arbitrary order (unlike what I am doing here). But when assembling `starts` and `counts` arrays, the assumption always seems to be that the dimensions are in that order. I can try to increase the flexibility a bit more by trying to mimic what TECA is doing with other dimensions. Though I do not fully understand all of that. Perhaps we can take a look together on Tuesday when we meet. You may know NetCDF conventions for climate data better than me and may have a better idea why TECA is doing things the way it is.

ghweber: I added a new commit that makes handling this in line with other dimensions. Dimension order now should not matter, but it still assumes a data layout of `ensemble`/`time`/'z'/'y'/'x'. Since TECA already did assume a layout of `time`/'z'/'y'/'x' this seems reasonable to me. Let me know if we should still do an issue.

io/teca_cf_layout_manager.cxx
```diff
@@ -293,18 +367,44 @@ int teca_cf_layout_manager::define(const teca_metadata &md_in,
         }
     }
 
+    std::string coord_array_name_in_file[4];
```
taobrienlbl: Could you please add some comments around/in this block?  It looks like this block looking for the presence of '/' to (1) determine if a coordinate is in a group and (2) extract the coordinate name itself if so.  This might not be obvious to others (or even to me) later on though

io/teca_cf_layout_manager.cxx
taobrienlbl: I see -- it looks like github is for some reason not showing me the most up-to-date version of this file.  I'm looking at the diff in visual studio code and see that you've removed this.

io/teca_cf_layout_manager.cxx
ghweber: That code doesn't exist anymore. I refactored things during a later commit, but it seems you are still seeing the old version. I added more comments in the function I created for this.

ghweber:
Superseded by #803

PR #804, opened 2025-01-31
Porting the DetectNodes/StitchNodes functions from TempestExtremes to TECA: CPU and GPU versions

amandasd:


amandasd:
@taobrienlbl Where should I include the contents of cuCompactor.cuh?

alg/cuCompactor.cuh
```diff
@@ -0,0 +1,165 @@
+/*
+ * cuCompactor.h
+ *
+ *  Created on: 21/mag/2015
+ *      Author: knotman
```
taobrienlbl: do we have permission to use this file here?  I don't see license info with this, and we need to make sure that anything we include is open source and has a license compatible with TECA's license

alg/cuCompactor.cuh
```diff
@@ -0,0 +1,165 @@
...
+
+__host__ __device__ int divup(int x, int y) { return x / y + (x % y ? 1 : 0); }
+
+__device__ __inline__ int pow2i (int e){
+	return 1<<e;
+}
+
+
+template <typename T,typename Predicate>
+__global__ void computeBlockCounts(T* d_input,int length,int*d_BlockCounts,Predicate predicate){
```
taobrienlbl: Could you please add some comments above each of these functions to briefly describe what the functions do?

alg/teca_detect_nodes.cxx
```diff
@@ -0,0 +1,1945 @@
...
+
+    if (fMinimum)
+    {
+       p_setMinima[q] = q;
+    }
+}
+// --------------------------------------------------------------------------
+template <typename T, typename T_SEARCH>
+__global__
+void find_all_local_maxima(
```
taobrienlbl: why are there two versions of this function (and of the _threshold variant of this function)?  Based on the way the're called later on the code, I'm thinking one is a GPU version and the other is a CPU version: is that right?  It might be preferable to update the names so that this distinction is more clear if so: e.g., `find_all_local_maxima_gpu()`

doc/rtd/applications.rst
```diff
@@ -1557,6 +1561,304 @@ Command Line Arguments
     displays both basic and advanced documentation together
 
 
+.. _teca_tempest_tc_detect:
+
+teca_tempest_tc_detect
```
taobrienlbl, line 1566: we should note in the documentation somewhere that this application is GPU-enabled

doc/rtd/applications.rst
```diff
@@ -1557,6 +1561,304 @@ Command Line Arguments
...
+Examples
+~~~~~~~~
+
+ERA5 data
+^^^^^^^^^
+
+.. code-block:: bash
+
+    #!/bin/bash
+    #SBATCH -C cpu
```
taobrienlbl, line 1837: can I suggest that this example be modified to (1) demonstrate running on the GPU, and (2) to replicate exactly the 2017 run that you did that we verified against Paul and Colins's TempestExtremes runs?

test/travis_ci/install_ubuntu_22_04.sh
```diff
@@ -45,3 +45,12 @@ pip3 install "numpy<2.0" mpi4py matplotlib torch
 
 # install data files.
 svn co svn://svn.code.sf.net/p/teca/TECA_data@${TECA_DATA_REVISION} TECA_data
+
+# install TELite library
+git clone https://github.com/LBL-EESA/TELite.git
```
taobrienlbl, line 50: could I request that you add installation of TELite to TECA superbuild?  https://github.com/lbl-eesa/teca_superbuild

taobrienlbl (approved):
This looks great Amanda!  I have a few minor comments to address; once they're addressed and once you think this is ready, feel free to merge!

CMakeLists.txt
```diff
@@ -281,6 +281,19 @@ if (NOT TECA_HAS_PARAVIEW)
...
 set(TECA_HAS_VTK ${tmp} CACHE BOOL "VTK features")
 
+# configure for TELite
+set(tmp OFF)
+find_package(TELite QUIET)
+if (TELite_FOUND AND ((DEFINED TECA_HAS_TELITE AND TECA_HAS_TELITE) OR (NOT DEFINED TECA_HAS_TELITE)))
+    message(STATUS "TELite features -- enabled")
+    set(tmp ON)
+elseif (REQUIRE_TELITE)
+    message(FATAL_ERROR "TELite features -- required but not found. set TELite_DIR to enable.")
```
taobrienlbl, line 292: I suggest that we set `REQUIRE_TELITE` to true by default; if a user does not want it, they can turn this flag off when building

doc/rtd/applications.rst
amandasd: (2) This is the exactly 2017 run that I did using default values. What is important here is how you will define the input file (era5_2017.mcf). (1) I can demonstrate using the GPU, but it is much slower than the CPU.

taobrienlbl: oh great - not sure what made me think this wasn't the same.  Thanks!  And good point about speed; we can add later once/if we have a PR focused on GPU performance.

doc/rtd/applications.rst
amandasd: I added the following in-bold sentence (in bold): "The cyclone detecton algorithm is based on two TempestExtremes codes :cite:`tempestextremes`: DetectNodes and StitchNodes. **It can be executed in parallel on CPUs and GPUs.** More details about the use of the application can ben found `here`_."

What do you think?

alg/cuCompactor.cuh
amandasd: Here is the license info: https://github.com/knotman90/cuStreamComp/blob/master/LICENSE
Is it compatible with TECA's license?

CMakeLists.txt
amandasd: Where should I set it to true by default?

taobrienlbl: not sure - where are some other defaults set in the cmake file?

amandasd: I didn't find any other defaults set in any cmake file.

taobrienlbl: not sure offhand then...though I know its possible

amandasd: Maybe in TECA_superbuild?

taobrienlbl: that's definitely a place to set it too. But for TECA itself, here's what I'm suggesting: https://stackoverflow.com/a/53159688

amandasd: Should it be here (line 132)?
Although I think REQUIRE_TELITE is always true, because if not, TECA will not compile.

```
# these control behavior of the build. most of the time you want to error
# out if these are not found. for those times when you don't set the corresponding
# REQUIRE variable to FALSE
set(REQUIRE_CUDA FALSE CACHE BOOL "Forces build failure when CUDA is missing")
set(REQUIRE_MPI TRUE CACHE BOOL "Forces build failure when MPI is missing")
set(REQUIRE_NETCDF TRUE CACHE BOOL "Forces build failure when NetCDF is missing")
set(REQUIRE_NETCDF_MPI TRUE CACHE BOOL "Forces build failure when NetCDF_MPI is missing")
set(REQUIRE_UDUNITS TRUE CACHE BOOL "Forces build failure when udunits2 is missing")
set(REQUIRE_BOOST TRUE CACHE BOOL "Forces build failure when boost is missing")
```

test/travis_ci/install_ubuntu_22_04.sh
amandasd: Should I clone/install TELite under `${TECA_PREFIX}`?

`${TECA_PREFIX:=/global/common/software/m1517/teca/perlmutter_${PARTITION}}`

CMakeLists.txt
taobrienlbl: yes, that looks right

test/travis_ci/install_ubuntu_22_04.sh
taobrienlbl: that might be a useful interim step, but I do think that adding it to the superbuild would be good; but that doesn't have to block this PR from merging.  I'll create another issue in teca_superbuild

alg/cuCompactor.cuh
amandasd: I just removed this file and commented everything related to its functions. Then we can uncomment it in the future and use it.

alg/cuCompactor.cuh
amandasd: We are no longer using these functions.

alg/teca_detect_nodes.cxx
amandasd: Done!

CMakeLists.txt
amandasd: Should we change anything in `docker/teca_cpu_debug/Dockerfile` and `setyp.py`?

taobrienlbl: oh, yes - almost certainly: good catch!  does TE Lite need to be pre-compiled, or is it just the source directory that's needed?

test/travis_ci/install_ubuntu_22_04.sh
amandasd: Thanks for creating an issue in teca_superbuild. Sorry, I was not clear in my question. My question is where I should clone/install TELite in `site/make_install_perlmutter_m1517.sh`? However, I have other questions about it because I think there are other changes to do in teca_superbuild.

CMakeLists.txt
amandasd: It needs to be installed (code below from `test/travis_ci/install_ubuntu_22_04.sh`)

`# install TELite library`
`git clone https://github.com/LBL-EESA/TELite.git`
`cd TELite`
`git checkout ${TECA_TELITE_REVISION}`
`mkdir build`
`cd build`
`cmake ..`
`make -j2 && make -j2 install`

test/travis_ci/install_ubuntu_22_04.sh
taobrienlbl: oh I see, yes, TECA_PREFIX

doc/rtd/applications.rst
taobrienlbl: great!

taobrienlbl:
> @taobrienlbl Where should I include the contents of cuCompactor.cuh?

is this question still needing an answer?

Amanda: No, it was deleted.

